Issue #40, opened 2015-01-13
Suggestion Feature. 'compose' method with deep extend.


raabbajam:
Can we use deep recursive copy / extend on 'compose' method?
Right now it doesn't. Can't we do this?
```javascript
var Validator = require('../libs/Validator');
// Validator has state.rules.flight = { ori: 'required|size:3', dst: 'required|size:3' }
// I want to create another factory with 
//  rules.flight = { ori: 'size:3', dst: 'required|size:3', rute: 'required' }; 
//  with 'ori' now is not required and 'rute' is required
describe('#compose with other states', function() {
		var _states = {
			rules: {
				flight: {
					ori: 'size:3',
                    rute: 'required',
				}
			}
		};
		var states = stampit()
			.state(_states);
		var Child = stampit.compose(Validator, states);
		it('should fails rules', function(done) {
            var query2 = {
                action: 'flight'
            };
			var child2 = Child({
				query: query2
			});
			var _query2 = child2.run();
            console.log(Validator.fixed.state.rules.flight); // { ori: 'required|size:3', dst: 'required|size:3' }
            console.log(Child.fixed.state.rules.flight); // { ori: 'size:3', rute: 'required' }
            // if we use deep recursive extend, rules.flight should be { ori: 'size:3', dst: 'required|size:3', rute: 'required' }
			expect(_query2.dst)
				.to.exist; // expecting query.dst to exist because it filled with error message that dst is still required
			done();
		});
```

on console:
``` 
1) Base validator #compose with other states should fails rules:
     AssertionError: expected undefined to exist
```

ericelliott:
I'm not sure I understand the question. Stampit already deep clones state.

```javascript
var b = { b: 'b' };
var foo = stampit().state({ a: b });
var bar = stampit().state({ c: b }); // => { a: { b: 'b' }, c: { b: 'b' } }

example.a.b = 'unique';
example; //=> { a: { b: 'unique' }, c: { b: 'b' } }
```

koresar:
@ericelliott you are wrong. If it does deep cloning then the following code should print '1 2'. But it prints '2 2'.
```js
var stamp = require('stampit')();
var state = { a: { b : null } };

state.a.b = 1;
var obj1 = stamp.create(state);
state.a.b = 2;
var obj2 = stamp.create(state);

console.log(obj1.a.b, obj2.a.b); // 2 2
```
[...]

PR #48, opened 2015-01-16
Mix in chain + some immutability

koresar:
@ericelliott please review carefully.

This is not the final PR. One more to go to reach full immutability.


koresar:
To simplify the review I will list changes:
 1. The `stamp.fixed` is not sharing anything with other stamps anymore. Plus unit tests for that.
 1. The `merge` and `mixin` functions of `mout` module as well as `mixinchain.js` function were replaced with enhanced (mostly copy&pasted) versions. Now we can control what's being copied/merged, and should it traverse prototype chain. Plus unit tests for that.
 1. The `Object.create` is reused (if present). Otherwise use the polyfil.
 1. The `methods` are now being copied from the given object **own** properties but not its prototype chain. Plus unit tests for that.
 1. `convertContructor` behaves exactly the same as in stampit 1.x regardless the changes inside the function. Plus unit tests for that.
 1. Unit tests were grouped (poorly).

I am ready to do any changes you wish.

mixer.js
```diff
@@ -0,0 +1,50 @@
...
+        loop(obj, copier, target);
+      }
+    }
+    return target;
+  }
+};
+
+module.exports.chain = true;
+module.exports.copyAll = null;
+module.exports.copyProp = function (val) {
```
ericelliott: Poorly named. This implies that this function makes a copy of something, when it's really testing whether or not it's a function.

mixer.js
```diff
@@ -0,0 +1,50 @@
...
+  return typeof val !== 'function';
+};
+module.exports.copyFunc = function (val) {
```
ericelliott: Also poorly named, and this is really just a duplicate of the previous function with reversed logic.

I suggest merging these into one function called `isFunction()`.

What's worse though is that none of these are used in this module at all. They should be defined in the module that uses them, instead.

mixer.js
```diff
@@ -0,0 +1,50 @@
...
+      obj = arguments[i];
+      if (obj) {
+        loop(obj, copier, target);
+      }
+    }
+    return target;
+  }
+};
+
+module.exports.chain = true;
```
ericelliott: The whole point of this branch is to make stamps immutable, but then you hang public mutable state on an exported object. This doesn't make any sense to me, and feels like a potential source of bugs and confusion.

stampit.js
```diff
@@ -8,17 +8,23 @@
...
 'use strict';
 var forEach = require('mout/array/forEach');
-var mixIn = require('mout/object/mixIn');
-var merge = require('mout/object/merge');
 var map = require('mout/array/map');
 var forOwn = require('mout/object/forOwn');
-var mixInChain = require('./mixinchain.js');
 var slice = [].slice;
 
+var mixer = require('./mixer');
```
ericelliott: Why do I have to look at 3 different files to figure out the logic you're setting up in 3 lines of code? =)

merger.js
```diff
@@ -0,0 +1,58 @@
...
+          },
+          target);
+      }
+    }
+
+    return target;
+  }
+};
+
+module.exports.chain = true;
```
ericelliott: This is all duplicated in `mixer.js`.

mixer.js
koresar: I didn't quite get what you mean. Mixer.js is not related to stampit.
Would you please rephrase that?

stampit.js
koresar: Probably because the variable names are confusing. Any suggestions?

merger.js
koresar: Half of it, or mostly. But not all. deepClone and isObject are used here. What's the suggestion?

koresar: Oh, by 'this' you mean `,chain`. Right?

ericelliott: Yeah, I mean the shared module exports.

koresar:
Ok, @ericelliott, check it out.

ericelliott:
Cool, I'll be out of touch for a couple weeks. I'll look at this when I get back. Thanks!

koresar:
Implemented stamp immutability. Additionally I removed the triple copy&paste. The `stamp.fixed` modification code was in the: `stamp()`, `compose()` and `methods()`/`state()`/`enclose()` functions. Not any more.

koresar:
@ericelliott I believe it's time to move on. Would you like to review the code?

NB: the PR destination branch is 'v2_0' not the `master'.

stampit.js
```diff
@@ -8,25 +8,30 @@
...
 
-create = function (o) {
-  if (arguments.length > 1) {
-    throw new Error('Object.create implementation only accepts the first parameter.');
-  }
-  function F() {}
+var mixer = require('./mixer');
+var mixIn = mixer.getMixin(); // Regular mixin function.
+var mixInFunctions = mixer.getMixin(isFunction); // mixin for functions only.
+var mixInChainFunctions = mixer.getMixin(isFunction, true); // mixin for functions including prototype chain.
```
ericelliott: This function signature is a [boolean trap](http://ariya.ofilabs.com/2011/08/hall-of-api-shame-boolean-trap.html). Possible solution: named options hash.

stampit.js
```diff
@@ -8,25 +8,30 @@
...
+}
 
-create = function (o) {
-  if (arguments.length > 1) {
-    throw new Error('Object.create implementation only accepts the first parameter.');
-  }
-  function F() {}
+var mixer = require('./mixer');
+var mixIn = mixer.getMixin(); // Regular mixin function.
+var mixInFunctions = mixer.getMixin(isFunction); // mixin for functions only.
```
ericelliott: This is overly complicated. I'd prefer it if you just export all the various types of functions you need. Keep the DOT (Do One Thing) principle in mind when you write modules. You DOT and DRY don't have to conflict. Each of the separate modules can share dependencies for code reuse. The point is that we shouldn't expose the complexity of configuring mixer functions to the main stampit module. It should be able to simply require the functions it needs by name.

For more on how to keep module interfaces simple, watch the talk [Simplicity Matters by Rich Hickey](https://www.youtube.com/watch?v=rI8tNMsozo0).

dist/stampit.js
```diff
@@ -710,9 +738,13 @@ isStamp = function isStamp(obj) {
  * @return {Function}             A composable stampit factory
  *                                (aka stamp).
  */
-convertConstructor = function convertConstructor(Constructor) {
-  return stampit().methods(Constructor.prototype).enclose(Constructor);
-};
+function convertConstructor(Constructor) {
+  var stamp = stampit();
+  mixInChainFunctions(stamp.fixed.methods, Constructor.prototype);
+  stamp.fixed.state = mergeChain(stamp.fixed.state, Constructor.prototype);
```
ericelliott: This makes the previous line redundant, since calls will find everything on the instance and never look at the prototype chain.

dist/stampit.js
```diff
@@ -710,9 +738,13 @@ isStamp = function isStamp(obj) {
  * @return {Function}             A composable stampit factory
  *                                (aka stamp).
  */
-convertConstructor = function convertConstructor(Constructor) {
-  return stampit().methods(Constructor.prototype).enclose(Constructor);
-};
+function convertConstructor(Constructor) {
```
ericelliott: Couldn't this whole thing be rewritten with no change in instance behavior as:

```js
  return stampit().state(Constructor.prototype).enclose(Constructor);
```

?

test/stampit-specs.js
```diff
@@ -17,63 +18,45 @@ test('.create()', function () {
...
 
 test('stampit().methods()', function () {
-  var obj = stampit().methods({
-    foo: function () {
-      return 'foo';
-    },
-    methodOverride: function () {
-      return false;
-    }
+  var obj = stampit({
```
ericelliott: Why did you change this so it no longer uses `.methods()` for the initial setup? You realize now, you're not simply testing `.methods()`? You're also testing the parameter passing, which means you may get a failure during the `.methods()` test which is actually caused by the parameter functionality...

test/stampit-specs.js
```diff
@@ -87,28 +70,36 @@ test('stampit().methods(a, b)', function () {
...
+  
   equal(obj.foo.bar, 'bar',
     'Should set default state.');
 });
 
 test('stampit().state()', function () {
-  var obj = stampit().state({
-    foo: {bar: 'bar'},
-    stateOverride: false
+  var obj = stampit(null, {
```
ericelliott: This is also now testing both the parameter form and the `.state()` form. Why are you introducing these test ambiguities?

test/stampit-specs.js
```diff
@@ -277,104 +345,58 @@ test('stampit.convertConstructor()', function () {
...
 
-  equal(u.bar(), 'bar',
-    'Should be able to add new methods with .compose()');
-
-  equal(u.baz, 'baz',
-    'Should be able to add new methods with .compose()');
-
-});
-
-test('stampit.compose() with inheritance', function () {
```
ericelliott: Why did you delete this test?

stampit.js
koresar: Thanks for the link. You are right.

stampit.js
koresar: Got it.

dist/stampit.js
koresar: I'll check it.

dist/stampit.js
koresar: The `convertConstructor` doesn't work for old school classes which inherit other classes. Moreover, if the `Constructor` has some non prototype properties then they are not copied over either.

Basically, the two new comprehensive `'stampit.convertConstructor()'` tests fail.

test/stampit-specs.js
koresar: Sorry. I see your point.

test/stampit-specs.js
koresar: +1

test/stampit-specs.js
koresar: I didn't. I expanded it to test many-many other cases. See explanation above.

dist/stampit.js
ericelliott: Note that I replaced `.methods()` with `.state()`.

Does `.state()` follow the prototype chain or only work for own properties?

koresar: Yeah. I noted. My comments above are related to your sample code.

Check it yourself :)

stampit.js
koresar: Great talk!

dist/stampit.js
koresar: You are correct. Also, the 'converted' objects are not lightweight. Is that okay?

ericelliott: What is the objective you're trying to accomplish by using a state stamp here?

Would this still be necessary if we stopped filtering `.methods()` to prevent state from being stored on the prototype?

Should we even be doing that?

stampit.js
ericelliott: :+1:

dist/stampit.js
koresar: You are looking at the wrong commit. Please, check the combined changes. Here is the link to the one:
https://github.com/ericelliott/stampit/pull/48/files#diff-f547c18e2953252cb96744460b29a7cfR205

koresar:
@ericelliott I have just pushed changes which address all your concerns.

koresar:
Please see few more commits. Improved DOT and DRY of stampit.js file.

Created two more tests to improve DOT as suggested. They cover the bug #40.

ericelliott:
Cool.. I'll check out the updated diff as soon as I can.

stampit.js
```diff
@@ -189,16 +191,21 @@ isStamp = function isStamp(obj) {
  * @return {Function}             A composable stampit factory
  *                                (aka stamp).
  */
-convertConstructor = function convertConstructor(Constructor) {
-  return stampit().methods(Constructor.prototype).enclose(Constructor);
-};
+function convertConstructor(Constructor) {
+  var stamp = stampit();
+  mixer.mixInChainFunctions(stamp.fixed.methods, Constructor.prototype);
+  mixer.mixIn(stamp.fixed.state, Constructor);
```
ericelliott: What's the point of this? Very few Constructor functions have properties attached to them, and even if they did, it would be more appropriate to copy them to the stamp function itself than it would be to copy it as instance state.

stampit.js
```diff
@@ -189,16 +191,21 @@ isStamp = function isStamp(obj) {
...
+  stamp.fixed.state = mixer.mergeChainObjects(stamp.fixed.state, Constructor.prototype);
```
ericelliott: Isn't it true that this will still override any method copied to `stamp.fixed.methods`? So this makes line 196 useless...

stampit.js
koresar: If it converts classic constructors partially then we have to rename the function to `convertSomePropertiesOfTheConstructorButNotAll`. :)))

People usually have no idea if a constructor has some static props. For example `mongoose.Model`. It's got them all.

Sample code:
```js
function Constructor() {}
Constructor.clone = function (constructorObject) { ...
```

What's your proposal?

stampit.js
koresar: My bad. It's a badly named function. `mergeChainObjects` should be probably named as `mergeChainNonFunctionProperties`.

It merges everything which is not functions.

stampit.js
koresar: Oh, I see your point! Sorry being so stupid. :)
I take my words back. The line 197 should be eliminated I think.

koresar: Removed the line. See latest commit.

I am so sorry for the time taken.

ericelliott:
There are lots of test failures in this branch. Can you double check that you're not using any ES5 features without polyfills, etc...?

Please feel free to use Mout or Lodash functions instead of ES5 features.

All tests are broken on:

* Internet Explorer 8
* iPhone 5

The last test is broken on Android 4

ericelliott:
I think the Sauce Connect config is broken, but you can test manually using Sauce Labs.

koresar:
I'll check it.

koresar:
iPhone [passes](https://assets.saucelabs.com/jobs/bf326f8026d04cf98d869b6754bb566a/0003screenshot.png) fine.

Android also fine. The last test is added by test driver. Not stampit related.

IE8 issue "Object doesn't support property or method" can't fix myself. The `Object.create` shim is used but tests still fail. Digging in.

koresar:
Ok, all fixed. The 'bind' function does not exist in IE8. That was my first ever IE8 developing experience.

I would need to join "IE8 haters" community now. Who's in charge?

PR #58, opened 2015-05-05
Replace mout with lodash

### LINKED ISSUE #51: Use lodash instead of mout
Since lodash v3.0 you can `require('lodash/object/merge')` exactly the same way as we do with mout.

Lodash is a bit faster, a bit smaller, uses a bit less memory.

Additionally, lodash has [`_.matches`](https://lodash.com/docs#matches) function which could be handy implementing #50.

Also lodash is under active development, have better online docs and JSDoc.

sethlivingston:
Was pretty easy except that mout mixIn === lodash assign. All unit tests passing. Fixes #51.


koresar:
:+1: Thank you!
I did the same for v2_0 branch. Didn't push yet.

koresar:
How is that possible that the PR is not mergeable? Which branch you coded it against @sethlivingston ?

koresar:
It takes me 1 minute to reimplement the PR, but 30 minutes to merge yours. :)

koresar:
Also a weird Github bug. The 3 commits from February got to this PR somehow...

koresar:
The size of `stampit.min.js` using `mout` in 6928 bytes atm. If `lodash` is used the size increases to 23000 bytes. @sethlivingston what's your opinion on that?

sethlivingston:
It's telling me that it can't be merged because I don't have write access. Sorry I don't have a more helpful answer on that.

Personally I would make lodash an external dependency, that way stampit only contains the stampit code itself. Those who use stampit will likely be bundling or including lodash through a different approach, anyway. Browserify lets you specify which libraries are external dependencies, and it will leave them out of the final .js file.

Does that help?

sethlivingston:
Regarding the three extra commits -- maybe there was some way to assign my PR to a branch of the same name (feature-lodash) in this repository? I'm not sure. Feel free to just take the changes and ditch the PR.

koresar:
You are correct.
I doubt there are many people who are referencing "stampit.min.js" directly anyways.

I'll close the PR and reimplement in "v2_0".

Thanks for the help!

Cheers.

sethlivingston:
Thanks, @koresar!

ahdinosaur:
if it makes sense, i recommend using the [modularized lodash modules](https://www.npmjs.com/browse/keyword/lodash-modularized) instead of the monolithic module, as that way it's possible to de-duplicate common lodash dependencies with other modules.

sethlivingston:
Wow I didn't know those existed... very cool. Not sure how to make that friendly for stampit users, though. For example, I'm already included the full lodash library in the app I'm building, so I don't need the function modules. I could see how that would be very attractive to other app developers, though.

sethlivingston:
As long as we're discussing options, there are two more (that could be combined):

1. Yank the implementations out of mout or lodash and put them in stampit. There are only six I think, and two of them are forIn and forOwn. Do the extra bytes justify the simplicity that comes with zero dependencies?
2. Build separate libraries (pseudo-names): stampit-standalone, stampit-lodash-external, stampit-lodash-functions-external, stampit-underscore..........

koresar:
1. Branch 2_0 is rather different already. There are lodash 11 functions being used. (BTW, you implemented the PR against 'master' but submitted the it for v2_0.)
2. This is a good idea. Would you create an issue on that please?

sethlivingston:
Oh duh, that explains it.

And yes, I will...

koresar:
Done with https://github.com/ericelliott/stampit/commit/346be0e13394540131f06db21dad9baa1dbb6b0d

Issue #64, opened 2015-05-22
Adding static methods/vars to a stampit constructor?


DylanPiercey:
Is there any convenience method to add statics methods to an object constructor or does this have to be handled outside of stampit?

troutowicz:
I don't think so. The way I have been doing this is:

```js
...
// create stamp
...
export default stampit.extend(someStamp, {
  // static methods/vars here
});
```

ericelliott:
@troutowicz nailed it. That's how I do it. :+1:

ericelliott:
A `.static()` convenience method would be a nice addition to the 1.x line. We should also pull it into 2.x, of course. Do either of you want to do a PR?

@koresar, you want to lead this? If @DylanPiercey or @troutowicz want to do a PR, do you have time to review & merge?

troutowicz:
@ericelliott I can work on the PR :)

ericelliott:
https://github.com/ericelliott/stampit/pull/65

ericelliott:
@troutowicz I've added you as a collaborator and assigned the issue to you. Feel free to reassign to @koresar when you're ready for code review.

koresar:
Hey @troutowicz, thanks for the PR.

I like the idea of having statics for stamp!

However, I probably should inform that we are going to have "refs" feature (see #60). Basically, these are object references added to each object instance (not stamp). Statics in the object so to say. May I wonder why you need statics for stamps but not object instances? I'd like to find how many people would need this feature.

Also, stampit main objective are objects, not stamps. I doubt people would use this feature extensively, if use at all. I believe modules (like stampit) should do single job and do it well.

Moreover, I feel like introducing a whole new feature which saves people 15 characters
`stampit.extend(someStamp, { ... });` is not worth the hassle around it. We would need to support it preferably forever.

Please, explain me where I'm confusing things. I believe you have stong need in this feature. Some more dicussion is really necessary.

Thanks @DylanPiercey for the idea! Cheers

troutowicz:
Hi @koresar.

So `refs` is just the new `state`, and `state` binds to the stamp instance. That feature has it's uses, but some libraries, like [React](https://github.com/facebook/react), look at the constructor object for [certain properties](https://facebook.github.io/react/blog/2015/01/27/react-v0.13.0-beta-1.html#es6-classes). And this is where a declarative `static` convenience method would be super cool.

At the end of most React components, you will find several `objects` being assigned to the `class` declared earlier in the component.

[...]

PR #65, opened 2015-05-23
stamp.statics() 2.0

troutowicz:
Relates to issue #64 

- [x] set using factory arg
- [x] set using factory method
- [x] set/override using factory method chaining
- [x] set/override using stamp composition

I'm not sure on the convenience method name, would `static` be better?

One use case that will fail is composition with a stamp that is extended with an object of functions.

```js
var stamp = stampit.extend(stampit(), {
  foo: function () {
    return 'foo';
  }
});
```

The `fixed.statics` property does not get set in this case, resulting in the `foo` method being undefined.


ericelliott:
I'd prefer the singular, `static` (short for "static properties").

We'll also want to add documentation. =)

koresar:
If we merge this then stampit v1 and v2 will have different fourth parameter. Moreover, this PR lacks tests. I'd implement this feature in v2 which I'm going to finish today. :)

troutowicz:
@koresar It was decided in #64 that v1 should only get the method form of this feature.

I can probably write up the tests in the next 24 hours, I got sidetracked with #66 today.

koresar:
I somehow missed the decision taking message.

So, method is the only thing implemented. The fourth parameter for `stampit()` will not be implemented. Is that correct?

Also README better be updated I reckon.

What if someone would like to create a static called "state" or "compose" or "methods"?
```js
var stamp = stampit().static({ compose: () => { console.log('hello'); });
stamp.compose ... ?
```
We should warn people that using reserved function names may and will lead to conflicts and unpredicted behaviours. Any ideas?

koresar:
One more thing.

You might want to consider `converConstructor` statics support. See the sample unit test:
```js
export class Counter extends React.Component {
  static defaultProps = { initialCount: 42 };
}

var stamp = stampit.convertConstructor(Counter);
assert(stamp.propTypes.initialCount === 42);
```

ericelliott:
> So, method is the only thing implemented. The fourth parameter for stampit() will not be implemented. Is that correct?

True for v1. For v2 I'd like to see all params collapsed into a single options arg, where `static` will be available.

ericelliott:
> You might want to consider converConstructor statics support. See the sample unit test:

:+1:

troutowicz:
From what I've experienced, stampit is not compatible with ES6 classes.

https://github.com/babel/babel/issues/682

ericelliott:
> stampit is not compatible with ES6 classes

Can you write a test that demonstrates the problem with Stampit, specifically?

troutowicz:
```js
test('stamp from ES6 class', (t) => {
  t.plan(1);

  class Foo {}
  const fooStamp = stampit.convertConstructor(Foo);

  t.ok(
    stampit.compose(fooStamp)(),
    'should create instance'
  );
});
```

`TypeError: Cannot call a class as a function`

ericelliott:
Could you file that as a bug in a new issue?

troutowicz:
Sure. #75

troutowicz:
I am going to create another PR for the 1.x branch. @ericelliott I know you prefer `static` over `statics`, however jshint will puke complaining about `static` and it being a reserved keyword. Do we disable jshint reserved keyword checks?

koresar:
I'll reimplement this in v2.0 branch after we merge #76 to v1, and #72 to v2. The idea and approach will be reused of course.

This PR is too incompatible. Easier to (re)implement than to merge. I can do that hopefully this week.  I'd close this PR for now. Are you alright with that @troutowicz?

Sorry I can't get free much time lately.

troutowicz:
I'll close it for you :)

koresar:
Cheers @troutowicz :)

Issue #71, opened 2015-05-24
Update README to reflect changes in v2.0


koresar:
A list of things to mention:
* [x] New aliases (`refs`, `init`).
* [x] New `props` method and fourth argument to `stampit()`. The way it works.
* [x] Immutable stamps (chaining always creates new stamp).
* [x] The way (init (aka enclose) works - `{instance, stamp, args}`.
* [x] Short list of breaking changes...
* [x] Two examples (`clone` and `getStamp`) from #62 as code samples.
* [x] New core stampit API (`stampit({methods, refs, init, props }`).
* [x] New `static` feature.
* ...

ericelliott:
:+1:

koresar:
Everyone, who has write access, are welcome to edit the new README. Here is the direct link:
https://github.com/ericelliott/stampit/edit/README_v2_0/README.md

I encourage you to do it because, as you might have noticed, my English skill sometimes lack proper structure.

unstoppablecarl:
I think there needs to be something explaining that `stampit().init()` can be used to enclose private data and as an initializer/constructor. I am not sure where this should go though.

koresar:
How about #81?

koresar:
Okay.

The PR #83 is ready to be reviewed and merged into v2. Waiting for Eric for a day.

Thanks you people for collaboration. It is so joyful to know I'm not alone. :) Cheers!

unstoppablecarl:
There is a lot of content needed to fully explain and document `stampit`. I think the current readme has too much for one file. 
I propose:
 - README.md description, table of contents type list of links to other info, very basic code examples with links to more in depth descriptions/examples of methods in other files or wiki pages
 - examples.md/wiki in depth examples and explanations of methods and specific cases see #81 
 - api.md the api section of the README.md

@koresar if you can get on gitter I would like to discuss.

koresar:
* Would you do that the way you see it, please?
* I'm okay with any number of .md files, but I'm strongly against dev. docs in wiki. I usually open `node_modules/module/README.md` instead of going only for API documentation. Wiki is good for  non-development info (like proposals, drafts, or history tracking).
* :+1:

ericelliott:
* fourth argument to stampit()

Didn't we decide that we're switching to an options object in 2.0?

I don't want 5 arguments in the API.

[...]

PR #72, opened 2015-05-24
Init making stampit flexible

koresar:
I didn't update README yet. Will be updated with #71.

@ericelliott, @LegitTalon, @troutowicz please review.

Cheers.


koresar:
Sorry rushing you up. I intend to finalize v2.0 with few days time. This PR is what holds us down. Does anybody want to take a look? Run tests? Browse the code? I would prefer to run this project collectively if possible.
Cheers

troutowicz:
@koresar Been busy, I can look over it later today.

koresar:
Thanks mate. The main changes to `stampit.js` are around line 131. The corresponding tests are in the `test/init.js`.

package.json
```diff
@@ -32,9 +32,9 @@
   },
   "scripts": {
     "start": "npm run hint && npm run build && npm run uglify",
-    "test": "browserify test/stampit-specs.js | tape-run",
+    "test": "browserify test/stampit-specs.js | node",
```
troutowicz: just make this:

`node test/stampit-specs.js`

koresar: true

troutowicz:
The tests are organized much nicer :+1:

koresar:
I felt like having "npm start" is kind of misleading for a utility module. So I replaced it with more useful "prepublish" hook to make sure none publishes untested code. :)

koresar:
Done.

JosephClay:
Awesome.

I'm running into a jshint error with `test/*`...seems to work passing `test` without the slash+wildcard.

`mixer.js` also has some lint errors.

unstoppablecarl:
Why is `mixer.mixIn` designed to need to use `.apply` couldn't it just use a param instead of `this` and be much more performant? 

https://github.com/ericelliott/stampit/blob/init-making-stampit-flexible/stampit.js#L54

JosephClay:
@unstoppablecarl Would be possible if the arguments were known for [`.mixIn`](https://github.com/ericelliott/stampit/blob/init-making-stampit-flexible/mixer.js#L25)..but I don't think it would be worth duplicating the function for an optimization.

unstoppablecarl:
What about using a faster `.apply` ?

Especially when not using any context.

https://github.com/codemix/fast.js/blob/master/function/applyNoContext.js

troutowicz:
I'm wondering the same @unstoppablecarl. What does the current implementation do differently than:

```js
function addRefs(fixed, states) {
  fixed.refs = fixed.state = mixer.mixIn(fixed.refs || fixed.state, states);
  return fixed.refs;
}
```
And why are we returning `fixed.refs`?

troutowicz:
Actually, is this not doing the same?

```js
function addRefs(fixed, states) {
  mixer.mixIn(fixed.refs || fixed.state, states);
}
```

JosephClay:
@troutowicz No, `objects` is being listed as a parameter, but is not being used. The function is looping through the arguments `Object.assign` style. see: https://github.com/ericelliott/stampit/blob/init-making-stampit-flexible/mixer.js#L25

ericelliott:
> When will you have source.fixed.state and not source.fixed.refs ?
The same for all the other || fallbacks ?

This is support for legacy stamps.

> objects is being listed as a parameter, but is not being used. The function is looping through the arguments Object.assign style. 

This can be fixed after upgrading stampit 2.0 to use ES6 using rest params. Instead of specifying an unused formal param, we can just use `...args` in the function signature, which will give us an array of all the objects.

troutowicz:
@JosephClay If you look at the mixin function, `objects` is never used... `arguments` is actually whats being used in the loop. And as far as I can see, stampit never passes in more than two objects (destination and source). Am I missing something?

troutowicz:
Ah, I see why we are setting `fixed.state` or `fixed.refs` then returning.... its for chaining.

ericelliott:
I wouldn't worry too much about nitpicky implementation details unless you spot something that's really slow. What you should do is:

1. Download it
2. Test & lint
3. Do all the tests pass?
4. Look at all the tests. Do they really cover what they're meant to cover? Do the tests make sense?
5. Look at the implementation. Does it look **correct**? (Note, correct and how you would implement it are two different things).

If you see something and there is clearly a better way, or a learning opportunity for the team, feel free to speak up.

Code reviews shouldn't slow everything down and lead people into bikeshed discussions. They should be a basic code quality protection, just so we all have a bit more peace of mind that the code does what we say it does.

The only thing that matters in software development is that the users love the software. Does it look like it would work and you would love to use it? =)

JosephClay:
@ericelliott Agreed

@troutowicz I think we're both saying the same thing

koresar:
I'll see what I can do about the comments above in 12 hours.

koresar:
> I'm running into a jshint error "cannot open test/*" with test/*...seems to work passing test without the slash+wildcard.

You were probably running old JSHint. No worries. I've fixed that with last commit. `npm run test/index.js`.

koresar:
> I'm wondering the same @unstoppablecarl. What does the current implementation do differently than:

All the proposed implementations above simply break tests. Try it. :)

The tests are very comprehensive, my friends. Very.

koresar:
> What about using a faster .apply ?
I'd prefer later stage optimizations. Premature optimizations are general evil. (I specialize on application performance for few years now.)

Let us think of performance when somebody hit bottleneck and claim it's because of stampit. ;)

Nevertheless, it would be awesome to see a performance optimizing PR after we release v2.0.

In fact, I did analyze performance of stampit. The only bottleneck is the `function factory()`. That's why you can find few `if` statements there. ;)

koresar:
Changes:
* Linting every JS file we wrote.
* Fix linting.
* Rename unusual `mixIn` to `mixin`. Leave alias for compatibility of course.
* Add comments to those weird `||` operators. Hopefully, this will help future code readers.

LegitTalon:
Good stuff. Great discussion, my only concern is documentation but it looks like #83 incapsulates that. :+1:

ericelliott:
Is there an `.assign` alias as well as `.mixin` and `.mixIn`?

koresar:
Yes, there should be one. Thanks for the reminder.

ericelliott:
Are there tests for the aliases? I didn't see them...

koresar:
There are no tests for `stampit.method` aliases as of yet. I wanted to add them though.

ericelliott:
:+1:

PR #76, opened 2015-05-24
stamp.static() 1.x

troutowicz:
More info at #65


troutowicz:
I still need to add the ability to inherit static properties during composition.

koresar:
Let me review this for you. :)

stampit.js
```diff
@@ -164,9 +176,13 @@ compose = function compose(factories) {
       if (source.fixed.enclose) {
         f.enclose = f.enclose.concat(source.fixed.enclose);
       }
+
+      if (source.fixed.statics) {
+        f.statics = mixInChain(f.statics, source.fixed.statics);
```
koresar: This will mix two objects *including* their prototype chain. Was that intentionally? I feel like `mixIn` would be safer and more predictable behavior.

koresar:
All good except for one little thing I'd like to clarify. As soon as we resolve that it should be fine to merge.

Could you please also add few usage samples to the README? Also, it'd be good to warn people over reserved stamp methods: create, methods, state, enclose, compose, and fixed.

stampit.js
troutowicz: Not intentional, it should be `mixin`.

troutowicz:
Sure thing.

stampit.js
koresar: :+1:

ericelliott:
> reserved stamp 'static' properties: create, methods, state, enclose, compose, and fixed

We should explicitly test for those and warn about conflicts.

koresar:
> We should explicitly test for those and warn about conflicts.
Which way of warning do you think of, Eliot?

ericelliott:
We shouldn't warn at all in the 1.x branch.

In the 2.x branch we should log to the console **in the debug build only**.

troutowicz:
@ericelliott @koresar How should static composition work? deep merge? override last in?

ericelliott:
@troutowicz For some use cases, deep merge makes sense. For others, shallow is better.

I personally use statics for things like type declarations (e.g. React `propTypes`). I prefer deep merge for `propTypes`.

What do you think, @koresar? Obviously this feature may be used for more than just React components. =)

troutowicz:
Agreed. Over at react-stampit, that is exactly what I proposed.

https://github.com/troutowicz/react-stampit/pull/2

troutowicz:
Added sample and more docs to README. All that is left is to determine how to handle static inheritance during composition and what to do about JSHints reserverd keyword error when trying to use `static` instead of `statics`.

ericelliott:
Turn off the JSHint reserved keyword error. It hasn't been needed since ES5, since reserved words are now allowed as property names without quotes.

ericelliott:
Ignoring the `propTypes` use case, I think `assign` behavior is probably what most people will expect.

koresar:
Should we split the feature to "staticProps" and "staticRefs"?

troutowicz:
I think that's overdoing it, and kind of confusing. All statics are props, props of the factor.

troutowicz:
@ericelliott @koresar Made the changes, cleaned up the commits. Should be ready to merge :)

koresar:
All good. I'd merge in 6 hours or so if everyone is okay with that.

ericelliott:
I'm OK with it.

troutowicz:
Great. I'll let it sit while I sleep and merge in the morning if no objections arise.

koresar:
@ericelliott, we need you to bump version to 1.2 and publish. And we're done here.

ericelliott:
Done! 1.2 is now available on npm.

koresar:
Thanks!

PR #78, opened 2015-05-27
Old node support removed

koresar:
I'm removing node.js 0.6 and 0.8 support. It's too time-consuming. Many modules do not support them as well.


koresar:
I accidentally created this PR against the wrong branch. So I'm closing it. :)

PR #79, opened 2015-05-27
Removing support of node.js 0.6, 0.8 and iojs_v1

koresar:
node.js 0.6 and 0.8 support is rather time consuming. Many modules we rely on do not support them. Thus, we won't support old technologies as well.

If someone need node 0.6 support then please submit a PR and make sure it's running against Travis.


troutowicz:
:+1:

ericelliott:
> I'm removing node.js 0.6 and 0.8 support.

:+1:

PR #82, opened 2015-05-28
added "what is a stamp?" section

unstoppablecarl:


koresar:
That's a great addition! Thanks. :)

Issue #57, opened 2015-04-24
Deep cloning state for v2.0


koresar:
1. New `.refs()` method for stamps which should shallow copy references.
1. `state()` should be deprecated (in docs, not in code yet) in favour of new `refs()`. `state()` should be v1.x compatible. `state()` will essentially be an alias to `refs()`.
1. New `props()` methods should deep clone the given object(s).

ericelliott:
:+1:

koresar:
You reckon we should add fourth parameter to the 'stampit' function to accept deep cloned "props" object? @ericelliott

ericelliott:
Yes.

koresar:
@ericelliott Should I add the "refs" to the prototype, or to the object itself (as it is now)?

I'm asking that because "props" are definitely are going to be merged to the object itself. But they might conflict with "refs". (I hope I'm clear here.)

koresar:
On the other hand, adding "state" values to prototype might break someone's code.

koresar:
Yeah, definitely the "state" (aka "refs") should go to each stamp. Discussion is over. :))))))

koresar:
One more thing I am hesitating about. The `stamp(obj)` should deep merge the `obj` or shallow mix in the properties?

koresar:
JFYI: It's shallow mixed-in in stampit@1.1.

poke @ericelliott

unstoppablecarl:
I think it should shallow merge. I don't think you would need to deep clone for init.

ericelliott:
> On the other hand, adding "state" values to prototype might break someone's code.

:+1: 

> The stamp(obj) should deep merge the obj or shallow mix in the properties?

I think shallow is better for 2 reasons: 

1. It's not a breaking change
2. It's probably less surprising behavior

unstoppablecarl:
Also: if you do need to make a clone you can do that before passing it in, or as part of your factory code. If it clones by default and you need to set by ref, you just can't.

koresar:
All the good reason. Thanks, guys.
I'd need to update PR #60 then. I'll ping you when it's ready for review.
Cheers.

koresar:
#60 was implemented and merged to v2_0 branch. Closing this issue.

Issue #50, opened 2015-03-03
New feature: Debug mode that warns in the console when stamp props collide.


ericelliott:
We could diff the object after a stamp is applied and display all the changes in the console.

@koresar - any interest in this for 2.0?

LegitTalon:
This seems like a useful feature. When reasoning about an application the developer can more easily keep track of unwanted side-effects. It might also be helpful to have an optional "strict" mode of sorts that **prevents** prop collision by throwing an error.

ericelliott:
I'm not sure we want a strict mode, because it would prevent the very common **defaults and overrides pattern**.

LegitTalon:
Could be a per stamp setting maybe to provide a sort of constant prop? I'm only conceptually speaking here, just thought it would be worth considering.

koresar:
Firstly, I believe all the new features should be introduced after we finalize the 2.0 branch (merge existing PR's, implement #41).

The proposed feature looks useful, although a rare developer would need it. Instead of implementing every feature in the world we could leave extensibility points. For example the three new upcoming [functions](https://github.com/ericelliott/stampit/blob/mixInChain/stampit.js#L52) `addMethods`, `addState`, and `addEnclose` are the very good candidate to be the first point. See possible code sample:
```js
var oldAddState = stampit._addState;
stampit._addState = function (fixed, states) {
  console.log(_.difference(fixed.state, _.merge.apply(null, [].concat(states))));
  fixed.state = _.merge.apply(null, [fixed.state].concat(states));
};
```

I haven't thought about the extensibility of stampit much. There got to be better ways to do it.

ericelliott:
I agree, RE: finalize the 2.0 branch.

I also agree about making Stampit extensible.

However, I also think that static analysis of stamps is essential to Stampit as a tool for building large applications, and a tool that could very easily tell you which stamps are likely to play nice with each other would be very useful, indeed. Integrating it into Stampit core so that you can easily interact with it in the console *as you're building the application* seems like a very valid use-case to me ...

And it would only take a few lines of code to do it.

koresar:
Would you please provide a code sample on how you see the
> easily tell you which stamps are likely to play nice with each other 

?

LegitTalon:
Rough prototype of how it might work

```javascript
function checkClashes(objOne, objTwo) {
  var props = Object.keys(objTwo)
  Object.keys(objOne).forEach(function(key) {
    if (props.indexOf(key) !== -1) {
[...]

PR #96, opened 2015-06-09
Stamp function accept refs

koresar:
@ericelliott @troutowicz @unstoppablecarl Please, review.

The main change is the object instantiation. See `function Factory()`.

Now all the references mixed into an object instance as is was before. But then the `props` are **safely** merged into the instance. "Safely" means that no refs data will be overriden, but props will be merged to the existing properties where possible. See the unit test of this PR.

In my opinion this is the very best approach for all existing user cases I have ever met so far.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ or by [downloading the latest release](https://github.com/ericelliott/stampit/re
 
  * Create factory functions (called stamps) which stamp out new objects. All of the new objects inherit all of the prescribed behavior.
 
- * Assign properties by passing references object to the stamp (factory function).
+ * Assign properties by passing a references object to the stamp (factory function).
 
  * Compose stamps together to create new stamps.
 
@@ -396,15 +396,15 @@ components that are passed in or composed.
  * @param  {Object} [options.refs] A map of property names and values to be mixed into each new object.
  * @param  {Object} [options.init] A closure (function) used to create private data and privileged methods.
  * @param  {Object} [options.props] An object to be deeply cloned into each newly stamped object.
- * @return {Function(refs)} factory A factory to produce objects.
- * @return {Function(refs)} factory.create Just like calling the factory function.
+ * @return {Function} factory A factory to produce objects.
+ * @return {Function} factory.create Just like calling the factory function.
  * @return {Object} factory.fixed An object map containing the stamp metadata.
- * @return {Function(methods)} factory.methods Add methods to the stamp. Chainable.
- * @return {Function(refs)} factory.refs Add references to the stamp. Chainable.
- * @return {Function(Function(context))} factory.init Add a closure which called on object instantiation. Chainable.
- * @return {Function(props)} factory.props Add deeply cloned properties to the produced objects. Chainable.
- * @return {Function(stamps} factory.compose Add stamp to stamp. Chainable.
- * @return {Function(statics)} factory.static Add properties to the factory object. Chainable.
+ * @return {Function} factory.methods Add methods to the stamp. Chainable.
+ * @return {Function} factory.refs Add references to the stamp. Chainable.
+ * @return {Function} factory.init Add a closure which called on object instantiation. Chainable.
+ * @return {Function} factory.props Add deeply cloned properties to the produced objects. Chainable.
+ * @return {Function} factory.compose Add stamp to stamp. Chainable.
+ * @return {Function} factory.static Add properties to the factory object. Chainable.
 
 
 ## The stamp object ##
@@ -469,7 +469,7 @@ MyStamp.create().originalStamp === MyStamp; // true
 ### stamp.props() ###
 
 Take n objects and deep merge them safely to the properties. Creates new stamp.
-Note: the merge algorythm will not change any existing `refs` data of a resulting object instance.
+Note: the merge algorithm will not change any existing `refs` data of a resulting object instance.
 * @return {Object} stamp  The new stamp based on the original `this` stamp.
 
 
@@ -488,7 +488,7 @@ Alias to `stamp([properties] [,arg2] [,arg3...])`.
 Just like calling `stamp()`, `stamp.create()` invokes the stamp
 and returns a new object instance. The first argument is an object
 containing properties you wish to set on the new objects.
-The properties are copied by reference using standard mixin/extend/assign algorythm.
+The properties are copied by reference using standard mixin/extend/assign algorithm.
 
 The remaining arguments are passed to all `.init()`
 functions. **WARNING** Avoid using two different `.init()`
--- a/README.md
+++ b/README.md
@@ -54,6 +54,8 @@ or by [downloading the latest release](https://github.com/ericelliott/stampit/re
 
  * Create factory functions (called stamps) which stamp out new objects. All of the new objects inherit all of the prescribed behavior.
 
+ * Assign properties by passing references object to the stamp (factory function).
+
  * Compose stamps together to create new stamps.
 
  * Inherit methods and default state.
@@ -61,9 +63,9 @@ or by [downloading the latest release](https://github.com/ericelliott/stampit/re
  * Supports composable private state and privileged methods.
 
  * References are copied across for each instance.
- 
+
  * Properties are deeply merged and cloned for each instance, so it won't be accidentally shared.
- 
+
  * Initializers are called for each new instance. Provides wide extensibility to stamp behavior.
 
  * For the curious - it's great for [learning about prototypal OO](http://ericleads.com/2013/02/fluent-javascript-three-different-kinds-of-prototypal-oo/). It mixes three major types of prototypes:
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -464,7 +468,8 @@ MyStamp.create().originalStamp === MyStamp; // true
 
 ### stamp.props() ###
 
-Take n objects and deep merge them to the properties. Creates new stamp.
+Take n objects and deep merge them safely to the properties. Creates new stamp.
+Note: the merge algorythm will not change any existing `refs` data of a resulting object instance.
```
troutowicz: algorithm

stampit.js
```diff
@@ -126,9 +125,9 @@ stampit = function stampit(options) {
     addStatic(fixed, options.static);
   }
 
-  var factory = function Factory(properties, args) {
-    properties = properties ? mixer.merge({}, fixed.props, properties) : deepClone(fixed.props);
-    var instance = mixer.mixin(create(fixed.methods), fixed.refs, properties); // props are taking over refs
+  var factory = function Factory(refs, args) {
+    var instance = mixer.mixin(create(fixed.methods), fixed.refs, refs);
+    mixer.mergeUnique(instance, fixed.props); // props are safely merged into refs
```
troutowicz: Would it be better to throw a merge conflict here instead of silently skipping the dupe keys?

README.md
```diff
@@ -54,6 +54,8 @@ or by [downloading the latest release](https://github.com/ericelliott/stampit/re
 
  * Create factory functions (called stamps) which stamp out new objects. All of the new objects inherit all of the prescribed behavior.
 
+ * Assign properties by passing references object to the stamp (factory function).
```
troutowicz: by passing references -> by passing a references

README.md
```diff
@@ -392,15 +396,15 @@ components that are passed in or composed.
...
- * @return {Function} factory.init Add a closure which called on object instantiation. Chainable.
- * @return {Function} factory.props Add deeply cloned properties to the produced objects. Chainable.
- * @return {Function} factory.compose Add stamp to stamp. Chainable.
- * @return {Function} factory.static Add properties to the factory object. Chainable.
+ * @return {Function(methods)} factory.methods Add methods to the stamp. Chainable.
+ * @return {Function(refs)} factory.refs Add references to the stamp. Chainable.
+ * @return {Function(Function(context))} factory.init Add a closure which called on object instantiation. Chainable.
+ * @return {Function(props)} factory.props Add deeply cloned properties to the produced objects. Chainable.
+ * @return {Function(stamps} factory.compose Add stamp to stamp. Chainable.
+ * @return {Function(statics)} factory.static Add properties to the factory object. Chainable.
```
troutowicz: I don't think these additions are needed. We only care about the data types of the returned values.

koresar: Yeah, there is a better way. Let me revert this for now.

stampit.js
koresar: Any kind of `throw` is too unexpected. Any other ideas/solutions?

koresar:
Fixed comments above. Thanks guys!

unstoppablecarl:
I don't think ignoring existing properties is a good idea.
 - It breaks backward compatibility
 - does not follow our consistent 'last in' behavior
 - I think most would be surprised to find properties ignored when creating an object instance with a stamp.

We need a name for this arg. It should be referred to as `settings` (or something else) to avoid confusing it with `props`, it is difficult to talk about it and compare it to `props` if it is also called `props`.

I am starting to question if the `settings` in `var instance = myStamp(settings)` should be merged at all. It may be making too many assumptions about usage. I would rather see it passed to the `init` functions and let the developer decide how it is used. `init({ args: args, instance: instance, stamp: factory, settings: settings }` (where settings is the `refs` arg above). I don't think we can make the same assumptions of purpose or intent with `settings` as we can with `refs`, `init`, `props`, and `static`.

I realize this would be a serious breaking change to backward compatibility and therefore probably not wise but I wanted to see what others thought. Would this make sense if backward compatibility were not an issue ?

I would like to see a way to circumvent the settings merge behavior. I have some ideas I will put in a PR.

koresar:
> I don't think ignoring existing properties is a good idea.

Wha "ignoring" are you referring to? The code does not ignore anything.

koresar:
@unstoppablecarl #50 might give you some insight maybe.

unstoppablecarl:
> "Safely" means that no refs data will be overridden, but props will be merged to the existing properties where possible

I assume this means properties matching `refs` will be ignored?

unstoppablecarl:
We discusses this here and agreed that the `settings` object should be shallow merged #57

koresar:
Please, do not use any new terminology to replace the existing. What do you mean by `settings`?

unstoppablecarl:
`var instance = myStamp(settings)` what should I call it to differentiate from `props`?

koresar:
These are `refs`. Please, check the code - the variable name reflects the usage.
Also check the changes to the README. It also says `refs` or `references` literally everywhere.

koresar:
>I assume this means properties matching refs will be ignored?

Nope, they will be not. "props" will be merged into an instantiated object which already has all the `refs` mixed in. Pseudo code:
 ```js
var instance = Object.create(...);
_.mixin(instance, refs);
_.merge(instance, props);
```

**Comment by unstoppablecarl** [comment]
I didn't completely understand the intended behavior. I get it now that I did some testing.

This is the only strange behavior I found but I think it is acceptable given the way js works.

```js
var stamp = stampit().props({
    connection: 'my connection',
});

var instance = stamp({connection: undefined});
console.log(instance); // { connection: 'my connection' }
```

stampit.js
ericelliott: > "no refs data will be overriden"

Should we merge props before adding the refs, and then allow refs to override props? What are the advantages or disadvantages?

koresar: Disadvantage: any ref object might receive new properties from the `fixed.props`.

Advantages: `stamp()` argument now serves both purposes - refs and props.
```js
var stamp = stampit.props({ url: { host: 'localhost' } });
var instance = stamp({ url: { port: 8080 } }); // deeply merged -> { url: { host: 'localhost', port: 8080 } }

var stamp = stampit.refs({ url: { host: 'localhost', port: 8080 } });
instance = stamp.refs({ method: 'GET' }); // shallwo mixed -> { method: 'GET', url: { host: 'localhost', port: 8080 } }
```

See my point? It's a win-win situation.

koresar: Edited the comment above. Please, review on the github website.

koresar:
I'll merge this tomorrow if no objections.

stampit.js
ericelliott: :+1:

test/props-safety.js
```diff
@@ -26,27 +26,22 @@ test('Stamp props deep cloned for object created', function (t) {
...
-  t.equal(o2.bar, 'bar');
-  t.equal(o2.baz, 'baz');
-  t.equal(o1.deep.foo, 'override');
-  t.equal(o1.deep.bar, 'bar');
-  t.equal(o1.deep.baz, 'baz');
-  t.equal(o2.deep.foo, 'override');
-  t.equal(o2.deep.bar, 'bar');
-  t.equal(o2.deep.baz, 'baz');
+  t.equal(o1.shallow1, 'leave me as is', 'A conflicting shallow reference must not be touched by props');
+  t.equal(o1.shallow2, 'merge me!', 'A non conflicting shallow reference must be merged form props');
```
ericelliott: Typo? Should be "merged from props."

koresar: :+1: fixed

koresar:
@unstoppablecarl The strange behavior you found can be easily fixed.

Your code:
```js
var stamp = stampit().props({
    connection: 'my connection',
});

var instance = stamp({connection: undefined});
console.log(instance); // { connection: 'my connection' }
```

Replacing `undefined` with `null` fixes it:
```js
var stamp = stampit().props({
    connection: 'my connection',
});

var instance = stamp({connection: null});
console.log(instance); // { connection: null }
```

koresar:
Since Eric has done his part I'm merging this in.

@ericelliott, please publish to NPM.

Meanwhile, I'll work on the Advanced Examples.

ericelliott:
```
var stamp = stampit().props({
    connection: 'my connection',
});

var instance = stamp({connection: undefined});
console.log(instance); // { connection: 'my connection' }
```

That is a confusing example. Normally you only want to override a property *if the property is set* in the override object. By passing in `undefined`, you're explicitly stating that *the property is not set*, so if there is a default value, as is the case here, then the default value should be used, which is exactly what happens.

As @koresar pointed out, if you want to override it with a null value, you should pass `null` instead of `undefined`. If you want to unset the key entirely, use an `.init()` function and the delete keyword, or compose from a stamp that doesn't contain the unwanted key in the first place. ;)

unstoppablecarl:
I understand all of that. As I said before I think it is acceptable given the way js works.

ericelliott:
:)

PR #109, opened 2015-06-14
Advanced examples

koresar:
Plentiful of tricks and tips. Please, correct my English.
Also, if something is unclear, please tell me! I want to make sure the text is easy to understand to as many people as possible.
Thanks!

P.S. What do you think if we create a new repo (and npm module?) with "cool stamps" like those in the examples? (Cloneable, EventEmittable, ForcedDefaults, etc.)


ericelliott:
cc @troutowicz let's get some eyes on these. =)

koresar:
Click this link to see how the .md looks like renderred: https://github.com/stampit-org/stampit/blob/advanced_examples/ADVANCED_EXAMPLES.md

ADVANCED_EXAMPLES.md
```diff
@@ -0,0 +1,409 @@
...
+- [Hacking stamps](#hacking-stamps)
+  - ["Default" properties](#default-properties)
+  - ["Default" properties as composable behavior](#default-properties-as-composable-behavior)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+
+## `object.getStamp()`. 2 ways.
+
+> Run the examples below for yourself:
```
ericelliott: Shouldn't this be under a "Run the Examples" header above this first example?

koresar: > above this first example

Where exactly?

ericelliott: Just before ## `object.getStamp()`. 2 ways.

ADVANCED_EXAMPLES.md
```diff
@@ -0,0 +1,409 @@
...
+
+First, let's assume you have a following stamp:
+```js
+var User = stampit();
+```
+
+### First way
+Just compose the following stamp to any other stamp.
+```js
+var SelfAware1 = stampit.init(function (ctx) { // the context, has `stamp` property
```
ericelliott: Use ES6?

```js
var SelfAware1 = stampit.init(function ({ stamp }) {
  this.getStamp = () => stamp;
});
```

troutowicz:
I'll sift through some of these as soon as I can :)

ADVANCED_EXAMPLES.md
koresar: The `node stampit/advanced-examples/self-aware.js` won't work then. :)
I wanted to make sure people can run examples without installing anything.

ADVANCED_EXAMPLES.md
koresar: I have to disagree on that.
1. Some might not even see the item `Run the Examples` in the table of contents. They might start reading from the `Memory efficient cloning`.
2. The `Run the Examples` is too broad. I wanted them to read example on this page, and then go to shell and launch the **specific** example. Not all of them.

ADVANCED_EXAMPLES.md
ericelliott: How hard is it to run `npm install`?

koresar: :) okay. Agree :)

unstoppablecarl:
I think examples should go in a different repo but that is just my preference.

ADVANCED_EXAMPLES.md
```diff
@@ -0,0 +1,409 @@
...
+```
+So, now every object instance returns the exact stamp it was built with. Nice!
+
+### Second way
+Another composable stamp which does the same but in a memory efficient way.
+It attaches the function to the `.prototype` of the objects, but not to each one.
+```js
+var SelfAware2 = stampit.init(function (ctx) { // the context, has `stamp` property
+  if (!ctx.stamp.fixed.methods.getStamp) { // Let's add the method only once.
+    var stamp = ctx.stamp; // We should reference only one object, otherwise it is memory leak risky
```
unstoppablecarl: "memory leak risky" is not good grammar. Can you explain what you mean by memory leak risky? also

unstoppablecarl:
In the clone example(s). I believe the way it is written you will loose which properties are `props` and `refs` it will treat everything as `refs` right?

ADVANCED_EXAMPLES.md
```diff
@@ -0,0 +1,409 @@
...
+
+### Memory efficient cloning
+
+Let's reimplement the `Cloneable` stamp so that the `clone()` function is not attached 
+to every object but to the prototype. This will save us a little bit of memory per object.
+```js
+var Cloneable = stampit.init(function (ctx) {
+  if (!ctx.stamp.fixed.methods.clone) { // check if prototype has the clone() method already
+    var stamp = ctx.stamp; // Avoiding too much data referenced and potential memory leaks
+    ctx.stamp.fixed.methods.clone = function () { return stamp(this) };
```
unstoppablecarl: Would it be more clear to use `ctx.instance` instead of `this`?

ADVANCED_EXAMPLES.md
```diff
@@ -0,0 +1,409 @@
...
+What if you can't create an object right now but have to retrieve data from a server or filesystem?
+
+To solve this we can make **any** stamp to return `Promise` instead of object itself.
+
+First, let's assume you have this stamp:
+```js
+var User = stampit.refs({ entityName: 'user' });
+```
+
+The following stamp should be composed *the last*, otherwise it won't work.
```
unstoppablecarl: 'the last' should be 'last.

ADVANCED_EXAMPLES.md
```diff
@@ -0,0 +1,409 @@
...
+$ git clone https://github.com/stampit-org/stampit.git
+$ node stampit/advanced-examples/prevalidate.js
+```
+
+For example you can prevalidate the object instance before a function call.
+
+This stamp separates the function definition and the prevalidation logic:
+```js
+var UserWithValidation = stampit.methods({
+  authorise: function () {
```
unstoppablecarl: authorise should be authorize.

ADVANCED_EXAMPLES.md
```diff
@@ -0,0 +1,409 @@
...
+
+## EventEmitter without inheritance (`convertConstructor`)
+
+> Run the examples below for yourself:
+```sh
+$ git clone https://github.com/stampit-org/stampit.git
+$ node stampit/advanced-examples/event-emitter.js
+```
+
+You can have a stamp which makes aby object an `EventEmitter` without inheriting from it.
```
unstoppablecarl: any not aby

ADVANCED_EXAMPLES.md
```diff
@@ -0,0 +1,409 @@
...
+```
+Will print `{ first: "John", last: "Doe" }`.
+
+As of stampit v2 the `convertConstructor` has limitations. It can't handle constructors with arguments. 
+
+------------------------------
+
+## Hacking stamps
+
+Each stamp has the property `fixed`. It's an object with 4 properties. It's used by stampit in that order:
```
unstoppablecarl: change 'in that order' to 'in the following order'

ADVANCED_EXAMPLES.md
```diff
@@ -0,0 +1,409 @@
...
+```js
+var ForcedDefaults = stampit.init(function (ctx) {
+  stampit.mixin(ctx.stamp.fixed.methods, this._enforcedDefaults);
+});
+```
+The stamp above will add all the `_enforcedDefaults` to the `.prototype`.
+
+Let's create user name and password enforcement.
+```js
+var DefaultUserCredentials = ForcedDefaults.refs({ _enforcedDefaults: { user: { name: "guest", password: "guest" } } });
```
unstoppablecarl: Shouldn't this be `ForcedDefaults.props` ?

ADVANCED_EXAMPLES.md
```diff
@@ -0,0 +1,409 @@
...
+console.log(instance2.data); // 2
+```
+Will print `2`. But let's delete this property from the instance.
+```js
+delete instance2.data;
+console.log(instance2.data); // 1
+```
+Will print `1`. The `data` was removed from the object instance, but not from its prototype.
+
+### "Default" properties as composable behavior
```
unstoppablecarl: Couldn't this be accomplished using `props` or `refs`?

ADVANCED_EXAMPLES.md
koresar: Done.

ADVANCED_EXAMPLES.md
koresar: I've updated the text significantly to address this confusion.

ADVANCED_EXAMPLES.md
koresar: :+1:

ADVANCED_EXAMPLES.md
koresar: :+1:

ADVANCED_EXAMPLES.md
koresar: :+1:

ADVANCED_EXAMPLES.md
koresar: Changed the examples to be ES6. There should be no such confusion. :) Thanks

ADVANCED_EXAMPLES.md
koresar: Removed that sentence.

koresar:
@unstoppablecarl @ericelliott @troutowicz 
This PR is ready to be merged. Please, review. Otherwise I'd merge it into `master` tomorrow.

Please, pay attention to the `stampit.js` and couple of tests were changes. A little history behind that:
* When I was implementing stampit v2 I decided to limit `fixed.methods` to contain functions only. So, I used the `mixer.mixinFunctions(fixed.methods, methods)`. Stampit v1 doesn't do that.
* However, while I was implementing these Advanced Examples I realized that limiting abilities is the worst idea. :) Within this PR I changed the code back to `mixer.mixin(fixed.methods, methods)`.

:tada: :smiley:

stampit.js
```diff
@@ -45,7 +45,7 @@ function extractFunctions(arg) {
 
 function addMethods(fixed, methods) {
   var args = [fixed.methods].concat(methods);
-  mixer.mixinFunctions.apply(null, args);
+  mixer.mixin.apply(null, args);
```
troutowicz: Why?

ADVANCED_EXAMPLES.md
```diff
@@ -0,0 +1,596 @@
...
+const SelfAwareUser1 = User.compose(SelfAware1);
+```
+Now, let's create a user and call the `.getStamp()`:
+```js
+const user1 = SelfAwareUser1();
+assert.strictEqual(user1.getStamp(), SelfAwareUser1); // All good
+```
+So, now every object instance returns the exact stamp it was built with. Nice!
+
+### Attach function to prototype (memeory efficient)
```
troutowicz: memeory -> memory

ADVANCED_EXAMPLES.md
```diff
@@ -0,0 +1,596 @@
...
+Another composable stamp which does the same but in a memory efficient way.
+It attaches the function to the `.prototype` of the objects, but not to each one.
+```js
+const SelfAware2 = stampit.init(({ instance, stamp }) => {
+  if (instance.getStamp) {
+    stamp.fixed.methods.getStamp = () => stamp;
+  }
+});
+```
+The `ctx.stamp.fixed` property contains stamp's internal data.
```
troutowicz: Where is ctx declared in this example?

ADVANCED_EXAMPLES.md
```diff
@@ -0,0 +1,596 @@
...
+assert.strictEqual(user1.getStamp(), SelfAwareUser1); // All good
+```
+So, now every object instance returns the exact stamp it was built with. Nice!
+
+### Attach function to prototype (memeory efficient)
+Another composable stamp which does the same but in a memory efficient way.
+It attaches the function to the `.prototype` of the objects, but not to each one.
+```js
+const SelfAware2 = stampit.init(({ instance, stamp }) => {
+  if (instance.getStamp) {
```
troutowicz: Why is this if statement used?

ADVANCED_EXAMPLES.md
```diff
@@ -0,0 +1,596 @@
...
+### Attach function to object
+Just compose the following stamp to any other stamp.
+```js
+const SelfAware1 = stampit.init(({ instance, stamp }) => {
+  instance.getStamp = () => stamp;
+});
+```
+Let's compose it with the `User` stamp from above:
+```js
+const SelfAwareUser1 = User.compose(SelfAware1);
```
troutowicz: I think for the doc, we should not append a number to variables, it makes for easier reading.

`SelfAwareUser1` -> `SelfAwareUser`
`user1` -> `user`

ADVANCED_EXAMPLES.md
```diff
@@ -0,0 +1,596 @@
...
+## Self cloneable objects - `instance.clone()`. 3 ways.
+
+> Run the examples below for yourself:
+```sh
+$ git clone https://github.com/stampit-org/stampit.git
+$ cd stampit && npm i babel -g
+$ babel-node advanced-examples/cloneable.js
+```
+
+This is simple stamp with a single method and a single state property `prefix`.
```
troutowicz: This is -> This is a

ADVANCED_EXAMPLES.md
```diff
@@ -0,0 +1,596 @@
...
+
+First, let's assume you have a following stamp:
+```js
+const User = stampit();
+```
+
+### Attach function to object
+Just compose the following stamp to any other stamp.
+```js
+const SelfAware1 = stampit.init(({ instance, stamp }) => {
```
troutowicz: No need for the extra parenthesis

`({ instance, stamp })` -> `{ instance, stamp }`

ADVANCED_EXAMPLES.md
```diff
@@ -0,0 +1,596 @@
...
+const user1 = SelfAwareUser1();
+assert.strictEqual(user1.getStamp(), SelfAwareUser1); // All good
+```
+So, now every object instance returns the exact stamp it was built with. Nice!
+
+### Attach function to prototype (memeory efficient)
+Another composable stamp which does the same but in a memory efficient way.
+It attaches the function to the `.prototype` of the objects, but not to each one.
+```js
+const SelfAware2 = stampit.init(({ instance, stamp }) => {
```
troutowicz: No need for the extra parenthesis

`({ instance, stamp })` -> `{ instance, stamp }`

ADVANCED_EXAMPLES.md
```diff
@@ -0,0 +1,596 @@
...
+```sh
+$ git clone https://github.com/stampit-org/stampit.git
+$ cd stampit && npm i babel -g
+$ babel-node advanced-examples/cloneable.js
+```
+
+This is simple stamp with a single method and a single state property `prefix`.
+```js
+const PrependLogger = stampit.methods({
+  log: function (obj) {
```
troutowicz: If we are going to make these examples ES6, lets use ES6 everywhere :)

`log(obj) {`

ADVANCED_EXAMPLES.md
```diff
@@ -0,0 +1,596 @@
...
+const originalLogger = PrependLogger();
+originalLogger.log('hello');
+```
+Prints `STDOUT: hello`
+
+### Attach method to each object instance
+
+Let's implement a stamp which allows **any** object to be safely cloned: 
+```js
+const Cloneable1 = stampit.init(({ instance, stamp }) => {
```
troutowicz: No need for the extra parenthesis

`({ instance, stamp })` -> `{ instance, stamp }`

ADVANCED_EXAMPLES.md
```diff
@@ -0,0 +1,596 @@
...
+OUT: hello
+OUT: hello
+```
+The `logger` and `loggerClone` work exactly the same. Woah! 
+
+### Bind method to each object instance
+
+This is how you can implement self cloning different: 
+```js
+const Cloneable2 = stampit.init(({ instance, stamp }) => {
```
troutowicz: No need for the extra parenthesis

`({ instance, stamp })` -> `{ instance, stamp }`

ADVANCED_EXAMPLES.md
```diff
@@ -0,0 +1,596 @@
...
+Let's reimplement the `Cloneable` stamp so that the `clone()` function is not attached 
+to every object but to the prototype. This will save us a little bit of memory per object.
+```js
+const Cloneable3 = stampit.init(({ instance, stamp }) => {
+  if (!stamp.clone) { // check if prototype is already has the clone() method
+    stamp.fixed.methods.clone = function () { return stamp(this); };
+  }
+});
+```
+The `ctx.stamp.fixed` property contains stamp's internal data.
```
troutowicz: For the same reasons as above:

`ctx.stamp.fixed` -> `stamp.fixed`

ADVANCED_EXAMPLES.md
```diff
@@ -0,0 +1,596 @@
...
+OUT: hello
+```
+Objects have the same state again. Awesome!
+
+### Attach function to prototype (memory efficient)
+
+Let's reimplement the `Cloneable` stamp so that the `clone()` function is not attached 
+to every object but to the prototype. This will save us a little bit of memory per object.
+```js
+const Cloneable3 = stampit.init(({ instance, stamp }) => {
```
troutowicz: `({ instance, stamp })` -> `{ instance, stamp }`

ADVANCED_EXAMPLES.md
```diff
@@ -0,0 +1,596 @@
...
+  if (!stamp.clone) { // check if prototype is already has the clone() method
```
troutowicz: `// check if prototype  already has the clone() method`

ADVANCED_EXAMPLES.md
```diff
@@ -0,0 +1,596 @@
...
+```js
+const User = stampit.refs({ entityName: 'user' });
+```
+
+The following stamp should be composed *last*, otherwise it won't work.
+```js
+const AsyncInitializable = stampit.refs({
+  db: { user: { getById() { return Promise.resolve({ name: { first: 'John', last: 'Snow' }}) } } } // mocking a DB
+}).methods({
+  getEntity: function(id) { // Gets id and return Promise which resolves into DB entity.
```
troutowicz: `getEntity(id) {`

ADVANCED_EXAMPLES.md
```diff
@@ -0,0 +1,596 @@
...
+------------------------------
+
+## Dependency injection tips
+
+Using the [node-dm](https://www.npmjs.com/package/node-dm) dependency management module you can
+receive preconfigured objects if you pass a stamp to it. It's possible because stamps are functions.
+
+Self printing behaviour. An object will log itself after being created.
+```js
+const PrintSelf = stampit.init(({ instance }) => {
```
troutowicz: `({ instance })` -> `{ instance }`

troutowicz:
I'm going to stop there for now. These examples are nice @koresar . :+1:

ADVANCED_EXAMPLES.md
ericelliott: :+1:

ericelliott:
Thanks for the excellent work @koresar & @troutowicz!

stampit.js
koresar: We can't be sure. And it's good! See my comment above. TL;DR - advanced examples require this change.

In fact, I'd add `proto` alias to the `methods` method (maybe someday).

ADVANCED_EXAMPLES.md
koresar: Good catch!

ADVANCED_EXAMPLES.md
koresar: Yup. Changing it now. Thanks!

ADVANCED_EXAMPLES.md
koresar: I've missed the `!` symbol. Thanks mate. Fixed and added comment.

ADVANCED_EXAMPLES.md
koresar: I agree. Can you take a look at the `advanced-examples/self-aware.js`?
As you can see there are two `SelfAwareUser`'s. Any ideas on how to rename those to be short, compilable, **and clear**?

ADVANCED_EXAMPLES.md
koresar: :+1:

ADVANCED_EXAMPLES.md
koresar: Wrong. :) I won't run without those. Try yourself.
`$ babel-node advanced-examples/self-aware.js`

Actually, I thought so too, that those parentheses are not needed. Until I ran it. :)

ADVANCED_EXAMPLES.md
koresar: Nope. Invalid ES6 syntax.

ADVANCED_EXAMPLES.md
koresar: Ha! Leftovers after I converted the file from ES5 to ES6. Thanks! Fixed all the occurences.

ADVANCED_EXAMPLES.md
koresar: Invalid ES6 syntax.

ADVANCED_EXAMPLES.md
koresar: Invalid ES6 syntax.

ADVANCED_EXAMPLES.md
koresar: :+1:

ADVANCED_EXAMPLES.md
koresar: Invalid ES6 syntax.

ADVANCED_EXAMPLES.md
koresar: :+1:

ADVANCED_EXAMPLES.md
koresar: :+1:

ADVANCED_EXAMPLES.md
koresar: Invalid ES6 syntax.

koresar:
@troutowicz Don't stop! Correct more! :) Please.

koresar:
Ok. Examples are ready to go. All issues corrected.

Merging today soon.

ADVANCED_EXAMPLES.md
troutowicz: I just meant that the variables should be renamed in this file, `ADVANCED_EXAMPLES.md`. This file is what explains the advanced use cases, it should be as clear as possible. The example files can be left as is I think.

koresar: :+1:

koresar: Check it out. And merge. :)

ADVANCED_EXAMPLES.md
troutowicz: I guess I expected that to just work heh. Oh well. \o/

unstoppablecarl:
>@unstoppablecarl @ericelliott @troutowicz 
This PR is ready to be merged. Please, review. Otherwise I'd merge it into master tomorrow.

>Please, pay attention to the stampit.js and couple of tests were changes. A little history behind that:

>When I was implementing stampit v2 I decided to limit fixed.methods to contain functions only. So, I used the mixer.mixinFunctions(fixed.methods, methods). Stampit v1 doesn't do that.
However, while I was implementing these Advanced Examples I realized that limiting the freadom is the worst idea. :) Within this PR I changed the code back to mixer.mixin(fixed.methods, methods).

I agree with this completely. I think calling it `methods` is misleading though. It seems the intent now is that `fixed.method` will be used as the object instance prototype via `Object.create(fixed.methods)` . Thoughts?

troutowicz:
@unstoppablecarl, `fixed.methods` has always been used as the object instance prototype. See [here](https://github.com/stampit-org/stampit/blob/master/src/stampit.js#L125).

I disagree with this change. The reason the property is named `methods` is because functions are the only data type safe to include on a prototype. (changes to a non-primitive may affect all instances with that prototype). We aren't limiting freedom, we are enforcing an API that will not create unexpected behavior.

ADVANCED_EXAMPLES.md
```diff
@@ -0,0 +1,594 @@
...
+```
+Let's compose it with our `User` stamp:
+```js
+const AsyncInitializableUser = User.compose(AsyncInitializable); // The stamp produces promises now.
+```
+Create object (ES6):
+```js
+const userEntity = AsyncInitializableUser({ id: '42' }).then(console.log);
+```
+A random stamp received the behaviour which creates objects asynchronously. OMG!
```
troutowicz: I see both spellings in this doc, lets stick with one.

behaviour -> behavior

ADVANCED_EXAMPLES.md
```diff
@@ -0,0 +1,594 @@
...
+
+------------------------------
+
+## Dependency injection tips
+
+Using the [node-dm](https://www.npmjs.com/package/node-dm) dependency management module you can
+receive preconfigured objects if you pass a stamp to it. It's possible because stamps are functions.
+
+Self printing behaviour. An object will log itself after being created.
```
troutowicz: behaviour -> behavior

ADVANCED_EXAMPLES.md
```diff
@@ -0,0 +1,594 @@
...
+  getById(id) {
+    return this.findById(id).then(this.convert.bind(this));
+  },
+
+  search(fields = {price: {from: 0, to: Infinity}}) {
+    return this.find({category: fields.categories, price: {gte: fields.price.from, lte: fields.price.to}})
+      .then(this.convert.bind(this));
+  }
+})
+  .compose(Converter, DbToApiCommodityConverter, MongoDb); // Adding the missing behavior
```
troutowicz: This belongs one line up.

unstoppablecarl:
@troutowicz You disagree with the change to allow non-functions to be set in `fixed.methods` ? or my assertion that `methods` is misleading if it is not limited to methods?

troutowicz:
I disagree with allowing non-functions in `fixed.methods`. :)

unstoppablecarl:
I can see both arguments. To be honest I don't feel that strongly one way or the other, but I believe if non-functions are allowed `methods` is not a good name.

unstoppablecarl:
Maybe we should have `methods` only allow methods, and add `proto` (suggested alias for methods by @koresar) allow all. ?

stampit.js
unstoppablecarl: Can you show which examples require it?

ADVANCED_EXAMPLES.md
ericelliott: :+1: 

[International standard = American English](http://www.openculture.com/2015/06/a-proportional-visualization-of-the-worlds-most-popular-languages.html)

ericelliott:
Non-method delegate prototype props is a code smell. We should only add `.proto()` if there is loud demand from users (so far there hasn't been).

Even if there is a surge in demand for it, I would insist that we include a strong warning against it in the docs.

advanced-examples/async-init.es6
```diff
@@ -0,0 +1,21 @@
+// jshint ignore: start
```
ericelliott: Can we change this filename to `.js`? ES6 is the current standard. =)

stampit.js
ericelliott: Which example needs this change? I'd like to take a close look at it, because using the proto for non-methods is a strong code smell. Maybe the example can be improved, instead of changing stampit behavior...

The original version of Stampit did not filter out non-functions. We trusted that the name `.methods()` was good enough.

ericelliott:
Let's get this merge-ready ASAP. I'd like to have these ready for the 2.0 announcement. =)

stampit.js
koresar: The `hacking.js` example prints `undefined`. Whereas it should print 
```
No credentials were given:  { name: 'guest', password: 'guest' }
Credentials were given:  { name: 'admin', password: '123' }
Reusing the default credentials:  { name: 'guest', password: 'guest' }
```

Please. fill free to push your ideas to this branch.

koresar:
> We aren't limiting freedom, we are enforcing an API that will not create unexpected behavior.

That's why I called the example **Hacking stamps**. Also, that's why the files is named **Advanced** examples. :)

I'd love to hear your ideas about how to make the `hacking.js` not to print `undefined`. ;)

advanced-examples/async-init.es6
koresar: The files was renamed to `delayed-instantiation.js` but the origin got back to the filesystem somehow. Removed it. thanks.

troutowicz:
@koresar If you are talking about the first example in `hacking.js` then I would say that example should not exist. No point in showing a 'hack' that is considered a code smell, not to mention changing stampit's internal behavior just to support a clever example.

koresar:
@troutowicz Truth is yours. I'll either remove or update the example if come up with something else.

advanced-examples/hacking.js
```diff
@@ -0,0 +1,37 @@
...
+delete instance2.data; // Deleting 'data' assigned to the instance.
+console.log(instance2.data); // 1 <- The .prototype.data is still there.
+
+
+// -----
+
+const UndeletableDefaults = stampit.static({
+  forceDefaults(defaults) { //creating new function in the stamp
+    const defaultRefs = this.fixed.static.defaultRefs || {}; // Taking existing defaults
+    stampit.mixin(defaultRefs, defaults); // add new defaults to it.
```
ericelliott: We should use `.assign()` in examples - it'll be familiar to people who know the ES6 `Object.assign()`.

advanced-examples/hacking.js
```diff
@@ -0,0 +1,37 @@
...
+console.log(instance.data); // 1
+const instance2 = Stamp({ data: 2 }); // Creating second object. It'll have property 'data' too.
+console.log(instance2.data); // 2
+delete instance2.data; // Deleting 'data' assigned to the instance.
+console.log(instance2.data); // 1 <- The .prototype.data is still there.
+
+
+// -----
+
+const UndeletableDefaults = stampit.static({
```
ericelliott: This whole example seems... dangerous? Could be a bug magnet.

advanced-examples/hacking.js
koresar: :+1:

advanced-examples/hacking.js
koresar: Yup. I'll remove it.

ericelliott: :+1:

koresar:
Done. Ready to merge this time?

Issue #91, opened 2015-06-02
Bring Stampit 2.x up to modern coding standards


ericelliott:
* [x] Convert to ES6
* [x] Add automatic validation in precommit and prepublish
* [x] Add other useful scripts

See [prod-module-boilerplate](https://github.com/ericelliott/prod-module-boilerplate) and [a real example](https://github.com/ericelliott/irecord).

koresar:
@troutowicz Take the first one! :)

ericelliott:
@troutowicz You might like that real example link. Click it. =)

troutowicz:
I'll take the first one. \o/

@ericelliott Looks like I need to take a crash course in RP ;)

ericelliott:
## Learn about Reactive Programming

### Video

* [Asynchronous Programming at Netflix](https://www.youtube.com/watch?v=gawmdhCNy-A) [Jafar Husain](https://twitter.com/jhusain)
* [What is Reactive Programming?](https://www.youtube.com/watch?v=dwP1TNXE6fc) [Jafar Husain](https://twitter.com/jhusain) explains reactive programming
* [Immutability: Putting The Dream Machine To Work](https://www.youtube.com/watch?v=SiFwRtCnxv4) by [David Nolen](https://twitter.com/swannodette)

### Reading

* [Introduction to Reactive Programming](https://gist.github.com/staltz/868e7e9bc2a7b8c1f754)
* [The Dao of Immutability](https://medium.com/javascript-scene/the-dao-of-immutability-9f91a70c88cd) The Way of the Functional Programmer
* [The General Theory of Reactivity](https://github.com/kriskowal/gtor) What is all this talk about reactive? Functional? Promises? This is the beginning of a reactive programming bible.
* [Reactive MVC and the Virtual DOM](http://futurice.com/blog/reactive-mvc-and-the-virtual-dom) Great read, even if you're not a React user.

tropet123456:
Is it possible to get this new version (v2) from Bower? and How?

koresar:
The only way to get the buggy v2 is to clone the repository. It's the
current master branch.

koresar:
@troutowicz It's the time. You can start converting stampit to ES6.

Please take a look at the link above: https://github.com/ericelliott/prod-module-boilerplate
The approach there suits us perfectly, IMO. Don't hesitate to strip parts you think we don't need. :)

troutowicz:
:grinning: Should be able to start this over the weekend.

[...]

PR #110, opened 2015-06-14
ES6

troutowicz:
#91

### DIFF
--- a/src/stampit.js
+++ b/src/stampit.js
@@ -6,80 +6,78 @@
  * Copyright (c) 2013 Eric Elliott
  * http://opensource.org/licenses/MIT
  **/
-'use strict';
-var forEach = require('lodash/collection/forEach');
-var map = require('lodash/collection/map');
-var forOwn = require('lodash/object/forOwn');
-var isFunction = require('lodash/lang/isFunction');
-var isArray = Array.isArray;
-var isObject = require('lodash/lang/isObject');
-var create = Object.create;
-var slice = require('lodash/array/slice');
-var mixer = require('supermixer');
+import forEach from 'lodash/collection/forEach';
+import forOwn from 'lodash/object/forOwn';
+import isFunction from 'lodash/lang/isFunction';
+import isObject from 'lodash/lang/isObject';
+import map from 'lodash/collection/map';
+import * as mixer from 'supermixer';
+import slice from 'lodash/array/slice';
+
+const create = Object.create;
+const isArray = Array.isArray;
 
 /* jshint -W024 */
 
 // Avoiding JSHist W003 violations.
-var stampit;
+let stampit;
 
-function extractFunctions(arg) {
-  if (isFunction(arg)) {
-    return map(slice(arguments), function (fn) {
+function extractFunctions(...args) {
+  if (isFunction(args[0])) {
+    return map(slice(args), fn => {
       if (isFunction(fn)) {
         return fn;
       }
     });
-  } else if (isObject(arg)) {
-    var arr = [];
-    forEach(slice(arguments), function (obj) {
-      forOwn(obj, function (fn) {
+  } else if (isObject(args[0])) {
+    let arr = [];
+    forEach(slice(args), obj => {
+      forOwn(obj, fn => {
         arr.push(fn);
       });
     });
     return arr;
-  } else if (isArray(arg)) {
-    return slice(arg);
+  } else if (isArray(args[0])) {
+    return slice(args[0]);
   }
   return [];
 }
 
-function addMethods(fixed, methods) {
-  var args = [fixed.methods].concat(methods);
-  mixer.mixinFunctions.apply(null, args);
+function addMethods(fixed, ...methods) {
+  const args = [fixed.methods].concat(methods);
+  mixer.mixinFunctions(...args);
   return fixed.methods;
 }
-function addRefs(fixed, refs) {
-  var args = [fixed.refs || fixed.state].concat(refs);
-  fixed.refs = fixed.state = mixer.mixin.apply(null, args);
+function addRefs(fixed, ...refs) {
+  const args = [fixed.refs || fixed.state].concat(refs);
+  fixed.refs = fixed.state = mixer.mixin(...args);
   return fixed.refs;
 }
-function addInit(fixed, inits) {
-  inits = isArray(inits) ? extractFunctions.apply(null, inits) : extractFunctions(inits);
+function addInit(fixed, ...inits) {
+  inits = extractFunctions(...inits);
   fixed.init = fixed.enclose = (fixed.init || fixed.enclose).concat(inits);
   return fixed.init;
 }
-function addProps(fixed, propses) {
-  var args = [fixed.props].concat(propses);
-  fixed.props = mixer.merge.apply(null, args);
+function addProps(fixed, ...propses) {
+  const args = [fixed.props].concat(propses);
+  fixed.props = mixer.merge(...args);
   return fixed.props;
 }
-function addStatic(fixed, statics) {
-  var args = [fixed.static].concat(statics);
-  fixed.static = mixer.mixin.apply(null, args);
+function addStatic(fixed, ...statics) {
+  const args = [fixed.static].concat(statics);
+  fixed.static = mixer.mixin(...args);
   return fixed.static;
 }
 
-function cloneAndExtend(fixed, extensionFunction, args) {
-  args = arguments.length > 3 ? slice(arguments, 2, arguments.length) : args;
-  var stamp = stampit(fixed);
-  extensionFunction(stamp.fixed, args);
+function cloneAndExtend(fixed, extensionFunction, ...args) {
+  let stamp = stampit(fixed);
+  extensionFunction(stamp.fixed, ...args);
   return stamp;
 }
 
-function compose(factories) {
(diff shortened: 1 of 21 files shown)


troutowicz:
Did a first pass for ES6 changes, will go over again soon.

src/stampit.js
```diff
@@ -6,80 +6,77 @@
...
+    forEach(args, obj => {
+      forOwn(obj, fn => {
         arr.push(fn);
       });
     });
     return arr;
-  } else if (isArray(arg)) {
-    return slice(arg);
+  } else if (isArray(args[0])) {
+    return args[0];
```
koresar: Should return args array, right?

troutowicz: If I interpreted it correctly, the way you had it before `arg` was the first argument passed in. Now, `args` is all possible arguments.

src/stampit.js
```diff
@@ -6,80 +6,77 @@
...
   }
   return [];
 }
 
-function addMethods(fixed, methods) {
-  var args = [fixed.methods].concat(methods);
-  mixer.mixinFunctions.apply(null, args);
+function addMethods(fixed, ...methods) {
```
koresar: The function can be shortened to a single line now, if I remember ES6 correctly .

troutowicz:
```js
var addMethods = (fixed, methods) => mixer.mixinFunctions(...[fixed.methods].concat(methods));
```
As long as things stay readable...

src/stampit.js
```diff
@@ -114,7 +111,7 @@ function compose(factories) {
  * @return {Function(statics)} factory.static Add properties to the stamp (not objects!). Chainable.
  */
 stampit = function stampit(options) {
-  var fixed = {methods: {}, refs: {}, init: [], props: {}, static: {}};
+  let fixed = {methods: {}, refs: {}, init: [], props: {}, static: {}};
```
koresar: Can this be "const"? What about the rest of the "let"s?

troutowicz: `const` means cannot be mutated after declaration.

src/stampit.js
```diff
@@ -207,10 +203,9 @@ stampit = function stampit(options) {
      * @param {[Function]|...Function} factories Stampit factories.
      * @return {Function} A new stampit factory composed from arguments.
      */
-    compose: function (factories) {
-      var args = isArray(factories) ? factories : slice(arguments);
-      args = [factory].concat(args);
-      return compose(args);
+    compose(...factories) {
+      let args = [factory].concat(factories);
```
koresar: What about "compose(factory, ...factories)"?

test/stampit-api.js
```diff
@@ -30,7 +30,7 @@ test('incorrect stampit({ refs }) args', function (t) {
   t.same(stampit({ refs: 42 }).fixed.refs, {});
   t.same(stampit({ refs: null }).fixed.refs, {});
   t.same(stampit({ refs: new RegExp() }).fixed.refs, {});
-  t.same(stampit({ refs: [42] }).fixed.refs, {});
+  t.same(stampit({ refs: [42] }).fixed.refs, { 0: 42 });
```
koresar: Please, change this and all the rest logics in the tests back. Your changes are wrong.

troutowicz: I don't think my logic is wrong. `supermixer` and other mixin/merge functions iterates arrays the same as objects. So why should we treat an array as something special here?

src/stampit.js
troutowicz: Why?

src/stampit.js
koresar: Yes, you are correct. Thanks. :)

src/stampit.js
koresar: No, I mean the other way. :)
```js
return mixer.mixinFunctions(fixed.methods, ...args);
```

src/stampit.js
```diff
@@ -6,80 +6,77 @@
...
-  mixer.mixinFunctions.apply(null, args);
+function addMethods(fixed, ...methods) {
+  const args = [fixed.methods].concat(methods);
+  mixer.mixinFunctions(...args);
   return fixed.methods;
 }
-function addRefs(fixed, refs) {
-  var args = [fixed.refs || fixed.state].concat(refs);
-  fixed.refs = fixed.state = mixer.mixin.apply(null, args);
+function addRefs(fixed, ...refs) {
```
koresar: same here:
```js
return fixed.refs = fixed.state = mixer. mixin(fixed.refs || fixed.state, ...args);
```

And in the rest of the `addXXX` functions.

src/stampit.js
troutowicz: :+1:

src/stampit.js
troutowicz: Yep.

src/stampit.js
koresar: I'm pretty sure `const` mean you cannot assign to this variable any more.

The following works well:
```js
const obj = { a: 1 };
obj.a = 2; // no errors :)
```
But the following is error:
```js
obj = {}; // ERROR!
```

troutowicz: Ah, yeah. I'll make the changes.

src/stampit.js
koresar: Shorter, and does exactly the same:
```js
compose(...factories) {
  return compose(factory, ...factories);
}
```

test/stampit-api.js
koresar: Oh, yeah. In this place it's correct. Pardon. I clicked the wrong "+".

src/stampit.js
troutowicz: Sorry, I read your question wrong. Yes, I am making this change to all relevant locations.

test/stampit-api.js
```diff
@@ -39,7 +39,7 @@ test('incorrect stampit({ init }) args', function (t) {
   t.same(stampit({ init: 42 }).fixed.init, []);
   t.same(stampit({ init: null }).fixed.init, []);
   t.same(stampit({ init: new RegExp() }).fixed.init, []);
-  t.same(stampit({ init: [42] }).fixed.init, []);
+  t.same(stampit({ init: [42] }).fixed.init, [42]);
```
koresar: Here is definitelyy a bug. The `init` array **must** contain functions only.

troutowicz: I'll take a closer look. :)

src/stampit.js
```diff
@@ -6,80 +6,76 @@
...
+    const arr = [];
+    forEach(args, obj => {
+      forOwn(obj, fn => {
+        if (isFunction(fn)) {
+          arr.push(fn);
+        }
       });
     });
     return arr;
-  } else if (isArray(arg)) {
```
koresar: I know that tests don't fail, but why you've removed this `if` case?

src/stampit.js
```diff
@@ -219,35 +210,36 @@ stampit = function stampit(options) {
...
-  var stamp = stampit();
-  mixer.mixinChainFunctions(stamp.fixed.methods, Constructor.prototype);
-  stamp.fixed.refs = stamp.fixed.state = mixer.mergeChainNonFunctions(stamp.fixed.refs, Constructor.prototype);
-  addInit(stamp.fixed, function (opts) {
-    return Constructor.apply(opts.instance, opts.args);
-  });
+  const stamp = stampit();
+  stamp.fixed.refs = stamp.fixed.state = mergeChainNonFunctions(stamp.fixed.refs, Constructor.prototype);
+
+  mixinChainFunctions(stamp.fixed.methods, Constructor.prototype);
```
koresar: You've changed refs and methods initialization order. But, frankly, you did it right. We should have had that order from the beginning. Thanks :+1:

src/stampit.js
```diff
@@ -312,14 +292,17 @@ module.exports = mixer.mixin(stampit, {
    * @param {...Object} source An object to copy properties from.
    * @returns {Object}
    */
-  mixin: mixer.mixin,
+  mixin,
+  extend: mixin,
+  mixIn: mixin,
+  assign: mixin,
```
koresar: :+1:

koresar:
Ok. It all looks good to me now. I'm okay to merge. I'd wait for other guys to review though.

PS: We'd need to replace jshint with eslint with a different PR someday.

src/stampit.js
troutowicz: `isObject` returns true for all types of objects, including arrays.

troutowicz:
Cool. Yeah, lets let the PR sit for a bit.

:+1:  eslint

src/stampit.js
koresar: That's right! You've improved this old function to better. Awesome!

src/stampit.js
```diff
@@ -6,80 +6,76 @@
...
+  mixinFunctions,
+} from 'supermixer';
+
+const create = Object.create;
 
 /* jshint -W024 */
 
 // Avoiding JSHist W003 violations.
-var stampit;
+let stampit;
```
unstoppablecarl: Just curious, does using `let` instead of `var` do anything here?

troutowicz: No. But `let` is basically a replacement for `var`, no real reason to use `var`.

ericelliott: Consider `var` deprecated. I can't think of a good reason to use `var` instead of `const` or `let`.

Prefer `const` whenever possible.

PR #120, opened 2015-06-21
Add react-stampit link to README

ericelliott:


troutowicz:
Nice!

koresar:
This is good to mentions react-stampit. But the placement of the text is rather too high. React is somewhat irrelevant to our main objective - the stamps. In my opinion, this text should go somewhere below "What is a Stamp" section.

koresar:
@troutowicz You were too fast to merge. :)

Nevertheless, should we move the React mention lower?

troutowicz:
I merged simply because everyone has pretty much agreed the README needs a rewrite. Figured just having the mention for now wherever was fine.

koresar:
I believe that making README worse just before making a large blog post about stampit is not wise. Can we at least minimize the bloating impact?

Just use **bold** instead of h3. See below.

-----

**React Users.** Stampit loves React. Check out [react-stampit](https://github.com/stampit-org/react-stampit) for composable components.

troutowicz:
I don't consider the placement bad to be honest. The placement should be close enough to the top and noticeable enough for anyone to see without having to scroll through API related docs.

But your change would probably be noticeable enough.

koresar:
@troutowicz see #124

PR #123, opened 2015-06-21
Eslint

koresar:
* Killed jshint.
* Applied eslint. Config was taken from [Airbnb popular repo](https://github.com/airbnb/javascript). Modified for stampit needs.
* Tests were converted to ES6. (That was hard!)


koresar:
Pushed one more commit for code coverage.

As I expected it's 100% :)

Usage:
`$ npm tun cov`
and maybe
`$ open coverage/index.html`

.eslintrc
```diff
@@ -0,0 +1,188 @@
+/*
+This is Aribnb widely used linting configuration. https://github.com/airbnb/javascript
```
ericelliott: Good enough to start. Eventually I want to use the [prod-module-boilerplate rules](https://github.com/ericelliott/prod-module-boilerplate/blob/master/.eslintrc) for all our projects. If there's something wrong with those rules, we should fix them.

I've already invested significant time in fine-tuning those rules. AirBnB's rules are trying to be a standard for the whole world. We don't need a standard for the whole world, we need a standard that enforces our own coding standards. =)

ericelliott:
:+1:

Issue #122, opened 2015-06-21
Simplifying README


koresar:
This might sound as not an important issue, but in my opinion this is the most important one. People come across stampit and move on because readme is too broad.

There are ways to do a better README.
* http://danhough.com/blog/nodejs-maintainer/
* https://github.com/zcei/standard-readme

It would be great if someone would draft a better README. Any ideas are very welcome. Any! Anyone?

## Checklist

* [x] [Restructure README doc](https://github.com/stampit-org/stampit/issues/122#issuecomment-118570103)
* [x] [Add pseudo code & remove confusing stuff](https://github.com/stampit-org/stampit/issues/122#issuecomment-118512009)
* [x] Link to [Fluent talk](https://vimeo.com/69255635)
* [x] Link to [The Two Pillars of JavaScript Part 1](https://medium.com/javascript-scene/the-two-pillars-of-javascript-ee6f3281e7f3)
* [ ] Link to [Common Misconceptions About Inheritance in JavaScript](https://medium.com/javascript-scene/common-misconceptions-about-inheritance-in-javascript-d5d9bab29b0a)
* [ ] In all examples, always use the `stampit.*` shortcuts when we're not passing a param.

unstoppablecarl:
I think these primary questions need to be answered at the start of the file as concisely as possible.

- What is stampit?
- What is a stamp?
- Why is this better than Classes?

ericelliott:
We need to answer "what is stampit?" and "why should I care?" in a short paragraph or two at the top of our README.

Then move on to the [https://github.com/zcei/standard-readme](Standard README) features.

I have no objection to moving the detailed API docs to individual files in the `docs` folder.

unstoppablecarl:
> moving the detailed API docs to individual files in the docs folder.
Agreed

unstoppablecarl:
> We need to answer "what is stampit?" and "why should I care?" in a short paragraph or two at the top of our README.

I think we should answer them as concisely as possible in 1 or 2 sentences at the top and elaborate later in the readme. Maybe a link in the short answers to the detailed explanations. I think the lack of conciseness at the start of the file is what is leading to the inability to get what stampit is after skimming the top.

koresar:
:+1: Conciseness and examples!

ericelliott:
* [ ] What is Stampit?
* [ ] Getting Started
  - One very basic example of install and use (a, b, compose)
  - Link to advanced examples
* [ ] What can it do for me?
[...]

PR #126, opened 2015-06-21
[#122] Remove bad contributing section

ericelliott:
A more detailed contributing section should be added in the docs folder later.

See https://github.com/stampit-org/stampit/issues/122


koresar:
Looks very reasonable to me. Cheers

PR #129, opened 2015-06-26
Stamps as promised

koresar:
What's been done:
* You can return `Promise`'s from the `init()` function(s). I.e.
```js
stampit.init(() => Promise.resolve({ newInstance: 'data' }))
  .create().then((myNewInstance) => {
    // do things with myNewInstance
  });
```
If any of the `init`'s return promise, then doesn't matter what, the stamp will return a promise too. See `test/promise-init.js`

* Coverage top up to 100%. Added few more tests for that. Removed a code which is impossible to hit.
* Removed the dependency on the lodash functions: `map` and `forOwn`. The `stampit.min.js` got shorter even though I've added Promises support. :)


README.md
```diff
@@ -121,7 +126,8 @@ a(); // Object -- so far so good.
 a().getA(); // "a"
 ```
 
-Yes. Got it. In both of these instances, we actually created a brand new object, and then immediately threw it away, because we didn't assign it to anything. Don't worry about that.
+Yes. Got it. In both of these instances, we actually created a brand new object, and then immediately threw it away, 
+because we didn't assign it to anything. Don't worry about that.
```
ericelliott: Why did you insert a line break here? All viewers that aren't brain-dead automatically wrap lines based on the window viewport width. Inserting arbitrary line breaks just makes the doc look funny when you view it in plaintext with different window widths...

src/stampit.js
```diff
@@ -21,38 +19,40 @@ import {
...
 }
 function addRefs(fixed, ...refs) {
-  fixed.refs = fixed.state = mixin(fixed.refs || fixed.state, ...refs);
+  fixed.refs = fixed.state = mixin(fixed.refs, ...refs);
   return fixed.refs;
 }
 function addInit(fixed, ...inits) {
   const extractedInits = extractFunctions(...inits);
-  fixed.init = fixed.enclose = (fixed.init || fixed.enclose).concat(extractedInits);
+  fixed.init = fixed.enclose = fixed.init.concat(extractedInits);
```
ericelliott:
```js
fixed.init = fixed.enclose = [fixed.init, ...extractedInits];
```

ericelliott:
I didn't see any major issues here. Just a couple minor nitpicks. Feel free to merge, or fix and merge, at your discretion. =)

:+1:

README.md
koresar: This is Markdown. The idea behind it is to be readable both as raw text and as a rendered page. Read first sentence here: http://daringfireball.net/projects/markdown/

A long single line treated the same as the multiple lines - as a single paragraph `<p></p>`. Unless there are two line breaks `\n\n`, which are treated as the paragraph end and a new paragraph start `</p><p>`. Read first sentences here: http://daringfireball.net/projects/markdown/syntax#p

Having a long 200 characters line is really hard to read raw (i.e. in an IDE). And most text editors do not have word wrapping enabled by default.

src/stampit.js
koresar: That doesn't work, Eric.
```
$ babel-node 
> let f = [1, 2];
[ 1, 2 ]
> let f2 = [3, 4];
[ 3, 4 ]
> [f, ...f2];
[ [ 1, 2 ], 3, 4 ]
```

But if you meant this
```js
ixed.init = fixed.enclose = [...fixed.init, ...extractedInits];
```
then I would avoid this form, because it trasplies very long and ugly to ES5 decreasing readability and performance.

koresar:
I'd wait for @troutowicz opinion before merging this PR.

ADVANCED_EXAMPLES.md
```diff
@@ -217,7 +217,8 @@ First, let's assume you have this stamp:
 const User = stampit.refs({ entityName: 'user' });
 ```
 
-The following stamp should be composed *last*, otherwise it won't work.
+When combined the following stamp will make any existing stamp to return promise instead of an object instance.
```
troutowicz: make any existing stamp to return promise
->
make any existing stamp return a promise

koresar: :+1: fixed

README.md
troutowicz: I don't find this necessary. It actually makes it confusing for me, makes it seem like new paragraphs which leads me to expect a change in topic.

src/stampit.js
troutowicz: No one is supposed to be reading the ES5. Does this decrease performance? If not I would say we should go the ES6 route whenever possible.

koresar: I'd stand for readability here. Explicit `contact` makes it clear for everyone. Whereas not every developers knows the ES6 syntax.

And also performance. :)

src/stampit.js
```diff
@@ -120,15 +120,60 @@ const stampit = function stampit(options) {
...
+        if (!isFunction(fn)) {
+          return; // not a function, do nothing.
+        }
+
+        // Check if we are in the async mode.
+        if (!nextPromise) {
+          // Call the init().
+          const callResult = fn.call(instance, {args, instance, stamp: factory});
+          if (!callResult) {
+            return; // nothing got returned. Proceed.
```
troutowicz: This comment is confusing. By definition, `return` prevents proceeding.

src/stampit.js
```diff
@@ -120,15 +120,60 @@ const stampit = function stampit(options) {
...
+          if (!isThenable(callResult)) {
+            instance = callResult; // stamp is synchronous so far.
+            return;
+          }
+
+          // This is the sync->async conversion point. Since now our factory will return a promise, not an object.
+          nextPromise = callResult;
+        } else {
+          // As long as one of the init() functions returned a promise, now our factory will 100% return promise too.
+          // Linking the init() functions into the promise chain.
```
troutowicz: I should not have to horizontally scroll to read comments. :(

koresar: I'll do something about it. Horizontal scrolling is developers mini-hell. :)

README.md
koresar: I'm not sure what you are talking about. Do you like horizontal scrolling?

We, and numerous other projects, have the entire .md files paragraphs mane of multiple lines. See the bottom of this file, for example.

Are we bikeshedding here? :)

src/stampit.js
```diff
@@ -120,15 +120,60 @@ const stampit = function stampit(options) {
...
+          nextPromise = nextPromise.then((newInstance) => {
```
troutowicz: We need to stay consistent with our arrow function formatting. The rest of the library excludes parenthesis around single parameters.

(newInstance) -> newInstance

This PR has a few other places doing this as well.

koresar: Fixed. :+1:

src/stampit.js
koresar: :+1: updated the text.

README.md
troutowicz: Well, implementing these changes in this unrelated PR could be considered bikeshedding. ;)

The formatting of the raw text does not need to be any different than the resulting rendered text IMO.

koresar: And it is not different. I made it less different!

I made this change to avoid horizontal scrolling when I read the .md file.

It's what Markdown was designed for - to be comfortable to read as both text and HTML.

src/stampit.js
troutowicz: Considering stampit is now an ES6 project, we shouldn't need to worry about what a developer does not know about ES6. But if performance is actually a noticeable thing here, then by all means stick with `concat`. :)

troutowicz:
Before this gets merged, I would like to strongly recommend that PRs with this many commits should get squashed into one commit before being merged. It makes master's commit history easier to follow.

README.md
troutowicz: OK :)

src/stampit.js
koresar: Yep. Each `...` becomes a `for` loop and a new array object.

koresar:
Let me squash it now.

koresar:
There is the only way to squash - this PR should be closed and a new one opened.

troutowicz:
Nah, you can still use this PR. Squash, then `git push -f origin stamps-as-promised`

koresar:
Oh, too late. :) Sorry. Next time I'll `-f`.

troutowicz:
I restored the branch, try it now :)

koresar:
Ha, now the PR gis disabled by GitHub. It says `The branch was force pushed or recreated`.

troutowicz:
Try now @koresar

src/stampit.js
troutowicz: Hmm, are you confusing spread and rest? `...` in this case is spread, which is just sugar for concat.

```js
let arr1 = [0, 1, 2];
let arr2 = [3, 4, 5];

let arr3 = [...arr1, ...arr2];
```

is equal to this ES5

```js
var arr1 = [0, 1, 2];
var arr2 = [3, 4, 5];

var arr3 = [].concat(arr1, arr2);
```

troutowicz:
Cleaned up PR over at #133

Issue #118, opened 2015-06-20
Dependency check


ericelliott:
Make sure all installed dependencies are declared.

[Example](https://github.com/ericelliott/prod-module-boilerplate/blob/master/package.json#L22) (`"missing"`, `"extra"` and `"ok"` tasks).

See [Upgrade Stampit quality control and security meta](https://github.com/stampit-org/stampit/issues/116)

koresar:
I'm not getting. What's that and what's the point?

ericelliott:
This issue is a check to make sure that all the dependencies you've installed locally also exist in `package.json`, so when other people try to install it, they get all of the required dependencies installed. It also checks for dependencies that are required in `package.json` that are not used in the code (which sometimes results in false positives, so we don't count those ones as errors).

koresar:
> when other people try to install it, they get all of the required dependencies installed.

Isn't it the way `npm` work by default?

> It also checks for dependencies that are required in package.json that are not used in the code

Do you think there is a possibility for that to happen in 100 lines code base of stampit? I doubt.

The dependency check worth doing for large projects. For a utility module like stampit it's an overhead.

ericelliott:
> Isn't it the way npm work by default?

Not if it's missing from `package.json`, which is what this checks for.

> Do you think there is a possibility for that to happen in 100 lines code base of stampit? I doubt.

Hoping to get all my projects using the same standards so we don't have to reinvent the wheel for every new project. Think on the scale of stampit-org, instead of just Stampit. =)

> For a utility module like stampit it's an overhead.

The dep check will run very fast, and only on Travis (since we won't make it a pre-commit requirement).

Sound fair?

koresar:
Yup. Sounds good.

koresar:
See #132

PR #132, opened 2015-06-27
New command `npm run check`, automatically runs on prepublish.

koresar:
See #118

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,13 +23,12 @@
     "url": "git@github.com:stampit-org/stampit.git"
   },
   "dependencies": {
-    "babel-eslint": "^3.1.19",
     "lodash": "^3.9.1",
     "supermixer": "^1.0.2"
   },
   "devDependencies": {
     "babel": "^5.5.8",
-    "babel-eslint": "^3.1.17",
+    "babel-eslint": "^3.1.19",
     "browserify": "^10.2.1",
     "dependency-check": "^2.5.0",
     "eslint": "^0.23.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -23,6 +23,7 @@
     "url": "git@github.com:stampit-org/stampit.git"
   },
   "dependencies": {
+    "babel-eslint": "^3.1.19",
```
ericelliott: Should this be in `devDependencies`?

koresar: Sometimes npm just can't do it right :)

package.json
```diff
@@ -41,12 +42,16 @@
...
     "test": "npm run lint && babel-node test/index.js",
     "clean": "rimraf dist/* && mkdir dist || true",
     "uglify": "uglifyjs dist/stampit.js -m -c warnings=false -o dist/stampit.min.js",
     "lint": "eslint src && eslint test",
-    "build": "npm run clean && babel src --out-dir dist && npm run uglify"
+    "build": "npm run clean && babel src --out-dir dist && npm run uglify",
+    "deps": "npm run deps:missing && npm run deps:extra",
+    "deps:missing": "dependency-check package.json",
+    "deps:extra": "dependency-check package.json --extra --no-dev --ignore",
+    "check": "npm run deps && npm outdated --depth 0"
```
troutowicz: What is the purpose of running `npm outdated` here. It won't prevent publishing afaik, looks to be just some extra info for the developer.

koresar: Yeah. It won't stop from prepublishing, but will print a nice colourful table.

Also, we might use this `npm run check` separately, without publishing.

troutowicz:
:+1:

Issue #138, opened 2015-06-29
Official package for meteor


yasinuslu:
It would be awesome if you guys would provide an official meteor package.

http://meteor.com/

koresar:
It would be even more awesome if you can guide us. :)
I know what Meteor is. Although, have zero experience with it. Do you have any Meteor experience?

koresar:
Ok, here is the guide: https://atmospherejs.com/i/publishing

Our package depends on two other packages. Will meteor grab those two from npmjs if any? @yasinuslu

yasinuslu:
Hi, there is an api to mark your npm dependencies but than the package wouldn't provide stampit on browser. I'm trying to implement a basic package for it, i'll write here when i complete it. Then we can discuss if it's a good way or not, how to automate it :) momentjs provides official package and here is their implementation https://github.com/moment/moment/tree/develop/meteor

yasinuslu:
I used browserify package to make it available on client. On server the package simply exports stampit to global from npm.

ericelliott:
What's wrong with [the existing atmosphere package?](https://github.com/richsilv/meteor-stampit)

ericelliott:
Oh, I see, the existing package depends on the `./dist` folder that no longer exists in the repository. Is there any way to update the existing package instead of modifying this one? This is the first time I've seen a Meteor package wrapper add things to the original npm package instead of creating a wrapper package for atmosphere.

ericelliott:
cc @philcockfield - Is there a better way to handle this?

koresar:
Eric, will you take a lead on this meteor issue please as you are our publish-master. 😀

yasinuslu:
@ericelliott I believe it's more reliable for people to see `stampitorg:stampit` :)

ericelliott:
> I believe it's more reliable for people to see stampitorg:stampit

I am open to that, but I have no interest in maintaining a meteor package. Are you volunteering?

ericelliott:
Move discussion to https://github.com/stampit-org/stampit-meteor/issues/1

PR #139, opened 2015-06-29
meteor package for stampit #138

yasinuslu:
I have implemented initial version of meteor package. Ref: #138

### DIFF
--- a/meteor/.npm/package/.gitignore
+++ b/meteor/.npm/package/.gitignore
@@ -1,0 +1,1 @@
+node_modules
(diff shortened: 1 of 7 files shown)


ericelliott:
I'm not convinced that this belongs in the Stampit repository. Maybe it should be in a separate repo that wraps this one?

koresar:
Agree on separate repo. Will you create one Eric? And maybe even give @yasinuslu the writing permission.

yasinuslu:
Wrapper packages for both bootstrap and moment (only official packages i recall) resides in the original repo so i thought it was logical.

https://github.com/twbs/bootstrap/blob/master/package.js
https://github.com/moment/moment/tree/master/meteor

Package doesn't use any files from this repository so there is no problem separating it.

ericelliott:
See [stampit-meteor](https://github.com/stampit-org/stampit-meteor)

PR #140, opened 2015-07-03
Tracing property overwrites

koresar:
RDD - Readme Driven Development. First demo commit.

Please, assess and comment. If the approach is okay I'll proceed the development.

### DIFF
--- a/README.md
+++ b/README.md
@@ -608,3 +608,17 @@ will probably clash with each other, producing very unexpected results.
 
   t.bar(); // 'bar'
 ```
+
+### stampit.tracePropertyOverwrites() ###
+
+Useful for debugging purposes. Will print logs when stampit overwrites any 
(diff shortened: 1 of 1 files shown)


ericelliott:
Personally, I don't care about property name collisions -- with the defaults/overrides pattern, that is a common and expected situation.

What I want to be notified about is if `propTypes` conflict when I pass stamps into `stampit.compose()`.

koresar:
That's something new to me.

Okay, again. How do you see an API? Without you answering this question I keen to abandon this task/feature. **I literally do not know what you guys want.** :)))))))

/cc @troutowicz

ericelliott:
=)

ericelliott:
Let's close this and open a new one later when requirements are fleshed out better. =)

Issue #150, opened 2015-07-09
Make it easy to find dist (build) files.


dwiyatci:
Link  prominently to all dist options in the README:

* [x] CDNJS
* [x] Bower (but recommend npm instead)

## Original message

Hi guys,

could we have consistency of always having the **dist file(s)** included for each release? I mean, in the latest release v2.1.0, for example, it doesn't have this `dist` folder which contains `stampit.js` and `stampit.min.js`.

Cheers

troutowicz:
The distributables are available through cdnjs, is there a particular reason you need them in the projects repo?

https://cdnjs.com/libraries/stampit

dwiyatci:
I checked the cdnjs earlier. Not all versions (including the latest) available there. No particular reason; just to keep my lib always updated to the latest.

troutowicz:
The distributables on cdnjs are supposed to be updated automatically at npm publish iirc. Looks like something is broken.

@nkbt any ideas?

bekos:
We could use the `dist` folder if we are using `stampit` through bower instead of cdnjs.

nkbt:
Maybe it takes some time for cdnjs to crawl and update packages?

ericelliott:
Questions:

* [ ] What is the current official release? (should this be in the README?)
* [ ] What is the current published release on npm?
* [ ] What is the latest build available on CDNJS?
* [ ] Do we link prominently to the CDNJS releases in the README?

sethlivingston:
Current release is **2.1.0** (https://github.com/stampit-org/stampit/releases).

Current published release on NPM is **2.1.0** (https://www.npmjs.com/package/stampit)

Latest available build on CDNJS is **2.0.3** (https://cdnjs.com/libraries/stampit)

We do **not** link prominently to CDNJS in the README.

We do need to provide a /dist/ folder, at the very least for bower. Bower was built to prevent developers from having to fetch files from all over the planet, including from CDNJS. IMHO it also just makes sense to have a committed copy of stampit.js and stampit.min.js included with every release we tag.

ericelliott:
> IMHO it also just makes sense to have a committed copy of stampit.js and stampit.min.js included with every release we tag.

Build artifacts are usually not checked into source control, but I also don't see a problem with this for small library projects

sethlivingston:
The Bower distribution is up and running with all of the releases: https://github.com/stampit-org/stampit-bower, and we have re-registered our Bower package to point there.

[...]

PR #152, opened 2015-07-13
Bower: restoring bower.json and the dist folder (#150)

sethlivingston:
- Restores `bower.json` to the repository.
- Restores the `dist/` folder to the respository.


ericelliott:
If we're going to do this, we should ignore the build directory in diffs.

### Ignore build directory in diffs

You can [ignore build files in diffs](https://github.com/gitlabhq/gitlabhq/issues/6916#issuecomment-62520577).

Create a `.gitattributes` file in the repo root:

```
dist    -diff
```

koresar:
Should we add the `.gitattributes` to `.npmignore`?

P.S. This is a ".ignore hell". :)

sethlivingston:
@koresar At least it's a one-time thing and not even something we need to automate. :)

ericelliott:
:+1:

sethlivingston:
Welp, we have another problem-o. Stampit depends on supermixer, which also does not support Bower. Further, both projects would really need to provide a UMD wrapped file to support Bower well.

Sooooo.... do we still want to do this?

troutowicz:
The distributables should be packaged with webpack, similar to react-stampit.

sethlivingston:
OK I'll take a look at that -- hold this PR for now, though.

troutowicz:
Here is the webpack config for react-stampit.

```js
var path = require('path');
var minify = process.env.MINIFY || false;

module.exports = {
  entry: {
    'react-stampit': ['./src/index'],
    'react-stampit-with-addons': ['./src/addons'],
  },

  output: {
    path: path.resolve('./dist'),
    filename: minify ? '[name].min.js' : '[name].js',
    library: 'stampit',
    libraryTarget: 'umd',
  },

  module: {
    loaders: [
      { test: /\.js$/, loader: 'babel-loader' },
    ],
  },
};
```

koresar:
1. We have "browserify" in devDependecies. Why not reuing it?
2. We should not store `dist/` files in the git repo.
3. The bower should be published with an npm command, like `npm run bower`. It should generate the bower-ready files, publish them.
4. The `dist/` files should be ingnored in both .gitignore and .npmignore.

sethlivingston:
1. In progress.
1. Bower retrieves the dist/ files from our GitHub repo; they have to be stored there. (1)
1. In progress.
1. In progress.

(1) If you like me and about 43 years old, then yes we were brought up to never commit output files (back then, large binaries) in a repo. Times have changed, though. Many depend on our output file being there, and it's just a small script, so there's really no harm. Another alternative is to handle it like John-David Dalton has for lodash: he has created separate GitHub projects for his distributions, but I don't think we need to.

sethlivingston:
OK, this commit includes:

* We were not using browserify, so I removed it in favor of webpack since we're already using it on the react side. Let me know if this is incorrect, but all of the checks worked out.
* `npm run bower` which packages stampit for the web (Bower) and puts it in the /dist folder.
* Added `npm run bower` to the `npm run prepublish` script.
* Added the /dist folder to `.npmignore`
* A few other supporting changes.

Bower retrieves everything it needs, on demand, from our GitHub repo. There is no need (or capability) to "publish" to Bower. These are the only requirements for being published on Bower:

1. `bower.json` is set up the way we want it.
1. Our distributables for Bower are in the repo.
1. We tag our releases--this is how Bower discovers available versions.

troutowicz:
I think the build step should be called `build-browser` or `build-umd`. The packaged file is not specific to bower. Also, the files should be named `stampit.js` and `stampit.min.js`, no need to include 'web' in the name.

I'm not sure about adding this build step to the `prepublish` step. The distributables need to be tracked with git, `prepublish` does not do that on it's own.

sethlivingston:
1. Let's do `build-web`. It's not a UMD package.
1. We do need different file names in the /dist folder, because they are different files. `stampit.js` can only be consumed by node modules. `stampit.web.js` can be consumed by browsers.
1. I can remove from the `prepublish` step and simply add it to the `build` step. As long as the `.web` files are in the /dist folder when we commit to the repo, we're good with Bower.

troutowicz:
Sorry, I forgot that the dist/ folder gets populated at prepublish time. The way react-stampit tackles this is using a lib/ folder for the node build during prepublish, and using the dist/ folder for the web build that is tracked by git. This allows the file to be named properly.

Is there a reason that the webpack build is not being packaged as umd?

sethlivingston:
Nobody uses AMD and we already have a proper nodejs build... so the only thing we need for browsers is build with "stampit" as a global.

troutowicz:
Our current nodejs build is just a transpiled version of src/. It isn't a really a 'build'. It does not package all src files into one distributable. The src just happens to be one file at the moment. Packaging with umd provides what the transpile step already does while ensuring a single file.

sethlivingston:
OK, is that what we want to do? Two UMD files in `/dist/stampit.js` and `/dist/stampit.min.js`?

troutowicz:
:+1:  for me, lets see what @ericelliott and @koresar have to say.

koresar:
I bed your pardon.
> The dist/ files should be ingnored in both .gitignore and .npmignore.

I meant not the entire directory, but just the newly generated files.

By putting the `dist` directory into the `.npmignore` we are breacking npm publishing process. The thing is stampit is traspiled from ES6->ES5 so that other projects could use the module. The ES5 files are stored in the `/dist` directory. See the `package.json` line:
> `  "main": "./dist/stampit.js",`

But knowing how bower works I understand why people are creating separate repositories for package management.

The `/dist/stampit.js` and `/dist/stampit.min.js` files are standard CommonJS files which are published to NPM. People expect them to remain CommonJS.

koresar:
I believe we should:
1) Create a new repo for bower packages.
2) Give the full control to @sethlivingston and full freedom of implementing/maintaining it. :)

How about that?

ericelliott:
> I believe we should:
> 1) Create a new repo for bower packages.
> 2) Give the full control to @sethlivingston and full freedom of implementing/maintaining it. :)

> How about that?

I'm OK with this solution as long as it gets auto-updated with git commit hooks. We've already had somebody set up Bower and then abandon it. If we offer it, we should make sure it's always current.

ericelliott:
> Two UMD files in /dist/stampit.js and /dist/stampit.min.js?

I'd also be OK with this solution.

sethlivingston:
I think with Bower going away (a lot slower than most think, mind you), a separate repo makes good sense. The current stampit-org/stampit is a "pure" node module; it would be nice to keep it that way.

I need 1) someone to set up stampit-org/stampit-bower (or somesuch), *or* 2) someone to give me permissions to do that, *or* 3) if you want to make this "third party", then I'll put it underneath my GitHub.

ericelliott:
> with Bower going away (a lot slower than most think...

I think the real story is how much npm's growth is leaving bower users behind. Stampit was definitely not the first project to drop Bower support. I think a lot of the Bower ecosystem will soon be looking at very outdated modules, and only the most popular will have mirror repos in the Bower ecosystem to stay current.

sethlivingston:
> I think a lot of the Bower ecosystem will soon be looking at very outdated modules, and only the most popular will have mirror repos in the Bower ecosystem to stay current.

I agree, and that will probably speed Bower's decline.

sethlivingston:
I think we are done with this PR - time for it to meet its spirit horse.

koresar:
Thanks @sethlivingston !
I'm closing this one. Have a good fun with the new repo. :)

PR #158, opened 2015-07-28
Real life example in readme

koresar:
Trying to achieve that sort of a readme: https://github.com/stampit-org/stampit/issues/122#issuecomment-118570103

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,8 +200,8 @@ var guestAuthorizer = ApiAuthorizion();
 var userAuthorizer = ApiAuthorization({name: 'Moses'}, '0E46A279');
 
 // Silly, but proves that everything is as it should be.
-guestAuthorizer.authorize('can-sync-data').then(console.log).catch(...); // prints "true"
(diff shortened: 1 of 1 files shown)


JosephClay:
ES6 looks good, but I find the silly example is more clear to me than the real life. Personal preference.

koresar:
@JosephClay Good to hear that! I appreciate your honesty. :)
Can you try and scrub your feelings regarding the two and tell us **why** silly is better?

koresar:
I'm merging this in 8 hours. :) Alright?

ericelliott:
I also think the silly example is more clear. The problem with a real life example is that people have a harder time imagining how to apply it to their own problems. Real life examples should stay in the advanced examples, IMO.

ericelliott:
I do like the ES6 updates, though. Want to remove the real life bit and just merge the ES6 upgrade?

troutowicz:
> Real life examples should stay in the advanced examples

Agreed. :)

koresar:
Ok. Sure.

koresar:
Looks like all good. :)

PR #170, opened 2015-09-20
Clear up `methods` example

tcrosen:
I could be wrong, but I think this example was meant to also demonstrate the use of a local method.


troutowicz:
Good catch.

koresar:
@tcrosen This proves that even if a PR is being twice reviewed one will always have issues.
Tahnk you for the fix Terry.

ericelliott:
Believe it or not: I write unit tests for all my example code when I write an article or a book. I think unit testing your docs is generally a good idea. ;)

koresar:
:+1:

PR #172, opened 2015-09-29
Fix README code samples by replacing `this` with `instance`

skaapgif:
I couldn't understand why the samples didn't want to work on my machine! After reading the API docs I realised they were wrong...

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ Basic questions like "how do I inherit privileged methods and private data?" and
 Let's answer both of these questions at the same time. First, we'll use a closure to create data privacy:
 
 ```js
-const a = stampit().init(function () {
+const a = stampit().init(() => {
   const priv = 'a';
   this.getA = () => {
     return priv;
@@ -135,7 +135,7 @@ because we didn't assign it to anything. Don't worry about that.
 Here's another:
 
 ```js
-const b = stampit().init(function () {
+const b = stampit().init(() => {
   const priv = 'b';
   this.getB = () => {
     return priv;
@@ -162,7 +162,7 @@ But that's boring. Let's see what else is on tap:
 
 ```js
 // Some more privileged methods, with some private data.
-const availability = stampit().init(() => {
+const availability = stampit().init(({instance}) => {
   var isOpen = false; // private
 
   instance.open = function open() {


README.md
```diff
@@ -162,7 +162,7 @@ But that's boring. Let's see what else is on tap:
 
 ```js
 // Some more privileged methods, with some private data.
-const availability = stampit().init(() => {
+const availability = stampit().init(({instance}) => {
```
koresar: Oops. That was a mistake. Thanks!

README.md
```diff
@@ -135,9 +135,9 @@ because we didn't assign it to anything. Don't worry about that.
 Here's another:
 
 ```js
-const b = stampit().init(function () {
+const b = stampit().init(({instance}) => {
```
koresar: This example should be working. I would rather leave it as is to showcase that every initializer is bound to `this`. Could you please revert this line and the change above back to initial state?
Thank you

skaapgif:
Which is the preferred way of using the API? If using `this` is preferred perhaps the other examples should be changed to also use it? Since the example which uses `instance` doesn't have any specific requirement for using `instance` instead of `this` it might only cause confusion?

koresar:
@skaapgif I wouldn't say that there is a prefferred way. Both ways work. We need to show them both in the examples.

P.S.
Although, for stampit v3 we are thinking of removing the `instance` support from the first agument, and move it to the second argument of the `init()` functions. :) Sorry for much confusion.

skaapgif:
I've changed those examples back to using `this`. I've kept my change in favour of using ES6 function declaration instead of ES5 since I think it's bad style to use both in the samp code sample (the anonymous function being declared as ES5 and `getA` being declared with ES6).

I haven't used stampit enough to run into every use case, but if there's no real advantage for using `instance` I would suggest that `this` becomes the official preferred way. Giving people two ways to do the same thing without any trade-offs or advantages just makes it harder to get started. If we want to show both ways of using it, it might be better to introduce that as a separate code sample so that each code sample in the readme only introduces one concept at a time. Just my 2c :+1:

koresar:
> it might be better to introduce that as a separate code sample so that each code sample in the readme only introduces one concept at a time.

100% agree. PR's are welcome. :)

koresar:
Thank you Rudolf.

PR #173, opened 2015-10-01
Change examples back to `this` for initializers

koresar:
It's better to follow one way of doing things. The  `{instance}` in initializers confuses people when mentioned.


troutowicz:
Agreed. :+1:

PR #176, opened 2015-10-10
Fixed a typo changed "form" to "from"

jonboylailam:
### Before

Creates a new object instance form the stamp.

### After

Creates a new object instance from the stamp.


ericelliott:
Thanks. =)

PR #209, opened 2016-05-31
Proper multi-environment build using rollup

### LINKED ISSUE #208: Proper build for various environments
Regarding closed issue #189 we might give it another try. Call me masochist, but I kinda do enjoy fiddling with various build tools and stuff :)

I have run into this article recently: https://medium.com/@tarkus/how-to-build-and-publish-es6-modules-today-with-babel-and-rollup-4426d9c7ca71#.jj0esfcdm. 

I am going to implement this soon for another library I am publishing, so if it proves itself, I am willing to apply same thing for stampit, if you like it.

FredyC:
Here we go, I am not expecting it would be perfect for the first time, so please try it and tell me what should be different. For convenience I've packed dist folder so you can see produced result.

[dist.zip](https://github.com/stampit-org/stampit/files/291430/dist.zip)

To summarize what's different here. There are 3 bundles here without any external dependencies included. These are meant for a further bundling in a different environments.

**stampit.mjs** - full featured ES6 including modules, essentially same thing as source file
**stampit.js** - ES6 without modules to be used in current NodeJS 6
**stampit.es5.js** - ES5 with required polyfills included, mainly for browserify & webpack 1.x

Lastly there is also **stampit.full.js** (+ minified) which is also ES5 compatible, but with all dependencies included. This is drop-in bundle for CDNJS and similar. Here is also reason why I had to change the source code and instead of

    import isFunction from 'lodash/isFunction';
    import isObject from 'lodash/isObject';

... there is usual `import _ from 'lodash';` now. Unfortunately lodash doesn't support true tree-shaking yet and doing it in current way means that lodash would be included in every bundle, even just those necessary functions. With little help of [babel-plugin-lodash](https://github.com/lodash/babel-plugin-lodash) this is handled more correctly.

I have also replaced use of `assign` from lodash with `Object.assign`.

Closes #208

### DIFF
--- a/buildconfig.env.example
+++ b/buildconfig.env.example
@@ -1,3 +1,0 @@
-export sauceport=5555
-export sauceuser=<yourusername>
-export saucekey=<yoursaucekey>
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,4 @@
+{
+  "presets": ["es2015", "stage-1"],
+  "plugins": ["transform-runtime"]
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,9 @@
 language: node_js
 node_js:
-  - "0.12"
   - "0.10"
-  - "iojs-2.0"
+  - "0.12"
   - "4.0"
-install:
-  - npm install
+  - "5.0"
+  - "6.0"
 script:
   - npm run check
--- a/build/build.js
+++ b/build/build.js
@@ -1,0 +1,86 @@
+/* eslint-disable no-console */
+
+import pkg from '../package.json';
+
+import rollup from 'rollup';
+import babel from 'rollup-plugin-babel';
+import nodeResolve from 'rollup-plugin-node-resolve';
+import commonJs from 'rollup-plugin-commonjs';
+import uglify from 'rollup-plugin-uglify';
+
+const moduleName = 'stampit';
+
+function execute() {
+  return Promise.all([
+    makeBundle(
+      { format: 'es6', ext: '.mjs' }
+    ),
+    makeBundle(
+      { format: 'cjs', ext: '.js' },
+    ),
+    makeBundle(
+      { format: 'cjs', ext: '.es5.js',
+        babelPresets: ['es2015-rollup'],
+        babelPlugins: ['transform-runtime']
+      }
+    ),
+    makeBundle(
+      { format: 'umd', ext: '.full.js',
+        babelPresets: ['es2015-rollup'],
+        babelPlugins: ['transform-runtime', 'lodash']
+      }
+    ),
+    makeBundle(
+      { format: 'umd', ext: '.full.min.js', minify: true,
+        babelPresets: ['es2015-rollup'],
+        babelPlugins: ['transform-runtime', 'lodash']
+      }
+    )
+  ]);
+}
+
+async function makeBundle(config) {
+  const isUMD = config.format === 'umd';
+
+  const inputConfig = {
+    entry: 'src/stampit.js',
+    plugins: [
+      babel({
+        babelrc: false,
+        exclude: 'node_modules/**',
+        presets: ['stage-1'].concat(config.babelPresets || []),
+        plugins: config.babelPlugins || [],
+        runtimeHelpers: isUMD,
+        externalHelpers: isUMD
+      }),
+      nodeResolve({ preferBuiltins: true, browser: isUMD }),
+      commonJs()
+    ]
+  };
+
+  if (!isUMD) {
+    inputConfig.external = Object.keys(pkg.dependencies);
+  }
+
+  if (config.minify) {
+    inputConfig.plugins.push(uglify());
+  }
+
+  const outputConfig = {
+    dest: `dist/${moduleName}${config.ext}`,
+    format: config.format,
+    sourceMap: !config.minify,
+    moduleName: moduleName,
+    exports: 'named'
+  };
+
+  const bundle = await rollup.rollup(inputConfig);
+  await bundle.write(outputConfig);
+  console.log('created', outputConfig.dest);
+}
+
+console.log('building...');
+
+execute()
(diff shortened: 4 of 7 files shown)


FredyC:
Do you think it's actually necessary to support iojs for travis check? It existed for a very brief moment, people either stayed on 0.12 or went to 4.0 afterwards.

koresar:
@FredyC yeah, I agree. We should drop io.js. It was added at the time when io.js was the newest coolest thing. :)

koresar:
Please, allow me some time to review the PR. Busy life Getting ready for the upcoming conference here in Sydney.

FredyC:
Sure no worry, it's obviously still kinda broken, so I wont stop till there are green lights :)

FredyC:
Alright, almost there, only 0.10 is failing because of the rollup-plugin-node-resolve is expecting native Promise. I will try to create issue there, but I have doubt's about it.

Unfortunately without that plugin there is problem of bundling dependencies...

```
Treating 'lodash/assign' as external dependency
Treating 'lodash/assign' as external dependency
Treating 'lodash/assign' as external dependency
Treating 'babel-runtime/core-js/object/values' as external dependency
Treating 'babel-runtime/core-js/object/values' as external dependency
Treating 'lodash/isObject' as external dependency
Treating 'lodash/isObject' as external dependency
Treating 'lodash/isObject' as external dependency
Treating 'babel-runtime/helpers/toConsumableArray' as external dependency
Treating 'babel-runtime/helpers/toConsumableArray' as external dependency
Treating 'lodash/isFunction' as external dependency
Treating 'lodash/isFunction' as external dependency
Treating 'lodash/isFunction' as external dependency
Treating 'lodash/assign' as external dependency
Treating 'lodash/assign' as external dependency
Treating 'lodash/isObject' as external dependency
Treating 'lodash/isObject' as external dependency
Treating 'lodash/isFunction' as external dependency
Treating 'lodash/isFunction' as external dependency
Treating 'stamp-specification' as external dependency
Treating 'stamp-specification' as external dependency
No name was provided for external module 'babel-runtime/core-js/object/values' in options.globals – guessing '_Object$values'
No name was provided for external module 'babel-runtime/helpers/toConsumableArray' in options.globals – guessing '_toConsumableArray'
No name was provided for external module 'lodash/assign' in options.globals – guessing 'assign'
No name was provided for external module 'lodash/isObject' in options.globals – guessing 'isObject'
No name was provided for external module 'lodash/isFunction' in options.globals – guessing 'isFunction'
No name was provided for external module 'stamp-specification' in options.globals – guessing 'compose'
No name was provided for external module 'babel-runtime/core-js/object/values' in options.globals – guessing '_Object$values'
No name was provided for external module 'babel-runtime/helpers/toConsumableArray' in options.globals – guessing '_toConsumableArray'
No name was provided for external module 'lodash/assign' in options.globals – guessing 'assign'
No name was provided for external module 'lodash/isObject' in options.globals – guessing 'isObject'
No name was provided for external module 'lodash/isFunction' in options.globals – guessing 'isFunction'
No name was provided for external module 'stamp-specification' in options.globals – guessing 'compose'
```

FredyC:
Alright, we are all green lights now. I have added some summary to the initial comment.

I am not entirely sure how to verify that those bundles are working correctly. I mean first ones are not that different from source, that should be just fine. The ES5 bundle is bit more tricky as it would need to run through some other bundler. I did some manual test of the Full bundle http://codepen.io/FredyC/pen/yJyYaR?editors=0010

You might notice that there is one another slight problem. Since we are exporting stampit as default alongside with `isComposable` & `isStamp`, consumer on ES5 environment will need to use `require('stampit').default`.  Is it actually necessary to export these two separately since they are already included in default export? Exporting only default means rollup can handle it properly. More about it [here](https://github.com/rollup/rollup/wiki/JavaScript-API#exports).

FredyC:
I am sorry @koresar, but based on rude behavior of @ericelliott I am closing this PR and I am explicitly against merging it without my consent.

I take this very personally and I don't want to help people who behave like master of everything and treat other disrespectfully.

ericelliott:
Maybe we should be able to optionally import `isComposable` and `isStamp` with:

```js
import isComposable from 'stampit/isComposable`;
import isStamp from 'stampit/isStamp';
```

That way we can have a single `stampit` export.

koresar:
@ericelliott good point. We should do it is a different PR.

buildconfig.env.example
```diff
@@ -1,3 +0,0 @@
-export sauceport=5555
```
koresar: This file exists as an example for SauceLabs frontend unit testing.
I'm happy to remove it though. Not used for 1 year now.

FredyC: Exactly reason why I removed it,there is no hidden magic in it that cannot be added when needed.

koresar:
I'm not sure what I did wrong, but stampit v2 full minified took 13K.
The stampit v3 full minified is 44K.

We are using `_.mergeWith` in both stampit v2 and v3. But stampit v2 uses lodash v3, and stampit v3 uses lodash v4.
Looks like lodash v4 is a bulky thing unlike lodash v3.

Overall, LGTM.

koresar:
Ok. I pulled your fork, solved the merge conflicts. Done!

FredyC:
I had it prepared already, it was just waiting for push, but thanks I guess. There was also one issue I have noticed that build is actually run twice, in `precheck` and then in `prepublish`. It's not big deal I guess, but I wanted to remove that, oh well.

Anyway reason for the size isn't in lodash, but in Babel polyfills in my opinion.

FredyC:
I am not yet sure if using that `babel-plugin-transform-runtime` is a good thing or not. It should be somewhat clever and include only polyfills that are actually used, but looking randomly in produced core, I see it included polyfill for `Array.includes` although it's not used anywhere in there. However without it things like `Object.values` or `Object.assign` would not work. Only other option is to reference these specifically through `core-js`.

Issue #68, opened 2015-05-24
Tests for old IE compatibility (SauceLab integration reincarnation)


koresar:
After we migrated tests to tape (see #63) we do not run them against IE8 on SauceLab. Should be returned somehow.

Discussion is welcome.

ericelliott:
I'm OK with dropping IE8 support in version 2, but version 1 still needs it.

koresar:
What about IE9?

ericelliott:
We should keep supporting IE9 in v2. There are still lots of people running old versions of Windows that don't support > IE9, unfortunately. );

koresar:
As long as IE 8, 9 and 10 are not supported anymore we should aim to test IE11, Chrome, Safary, and FF.

FredyC:
I agree with @koresar. People running software that no longer has any support are just shooting themselves in the leg.

Btw, don't forget about the Edge. If you don't know about it, it's a Microsoft browser in Windows 10 with much faster release cycles. They are are almost done implementing whole ES2015 in it, even modules.

ericelliott:
Yep. IMO, our new targets should be IE11+ and all evergreens.

koresar:
#326 contains significant simplification of stampit module code. Thus, I was able to make unit tests runnable in browsers.
No automatic tests though.

koresar:
Implemented browser tests in v4. Manual tested in IE11. All good.

If someone would like to integrate stampit with SauceLab or similar - a PR is welcome.

PR #210, opened 2016-05-31
[WIP] Use ava instead of tape for unit tests

FredyC:
There is no big reason behind it except it's just modern tool. It's based on tape so writing tests isn't much different except some assertions are named differently. No need to specify `t.end()` for each test anymore.

It runs all tests in parallel so it's definitely faster to execute. It also has very useful watch mode which runs tests only for changed files so there is even more time saved during development.

Since there will be conflicts between this and PR #209 let's do that first and then I will update this.


ericelliott:
Ava doesn't work in browsers, yet.

FredyC:
@ericelliott  I don't see any browser tests in stampit, so how does that relate?

ericelliott:
Apologies for being so short. I'm not trying to be rude, I just have limited time to review PRs between my other responsibilities. I shut this down quickly because Ava is a non-starter for browser testing, and our tests must be browser compatible so that we get get the browser tests running again. Stampit used to have browser tests that ran in the cloud on Sauce Labs, and we need to make sure we can easily do that again.

koresar:
Hey, Daniel. Thank you so much for the initiative. I believe ava is a better tool than tape.
Same time Eric is correct. We have plans to revive the browser tests. Here is the long living opened issue for it: https://github.com/stampit-org/stampit/issues/68

I really feel this PR is a great coding art. Tomorrow I'll investigate the ava usage inside browsers, there might be workarounds (via polyfills, or PhantomJS, or else). Will do my best and come up with more info sometime Friday or Saturday.

FredyC:
I don't see reason why it could not run in browser as well. If the problem is running in parallel, it possible to change it to serial execution. Otherwise I cannot imagine what would be difference.

koresar:
There was a time when stampit tests were failing in IE8. Simply because i used a language construct not available in IE8. IE11 could become the new IE8.

As you suggested earlier, stampit will support Promises in the future. But [IE11 and old Android <4.4.4 Browser do not support Promises](http://caniuse.com/#search=promise) natively.

It's just one of the possible scenarios. There might be other things, like iOS Safary, IE11, and Android browsers <5.0 do not support [arrow](http://caniuse.com/#search=arrow) functions.
etc. etc. etc.

I wish I would ditch that IE11 and Android old crap. But people rely on us.

ericelliott:
Getting Ava to work in browsers is not a small task. There is a [long open Ava issue tracking the slow, not-a-top-priority progress](https://github.com/avajs/ava/issues/24).

FredyC:
Ok I see, but I believe that solution is simply running browser tests against bundle that is made to be executed in browser. With my rollup changes (#209) this should be a breeze.

ericelliott:
I put a little time into trying to get browser tests working for stamp-utils. I felt like I was close, but I ran out of time to work on it. Getting the compose tests running from `check-compose` seems like the next logical step. =)

ericelliott:
> Ok I see, but I believe that solution is simply running browser tests against bundle that is made to be executed in browser. With my rollup changes (#209) this should be a breeze.

Yep! I was looking at doing the same thing with `stamp-utils` yesterday.

ericelliott:
Sadly, getting Ava to work would be a lot harder than rolling a bundle. Ava uses a bunch of Node features that aren't available in browsers. Exploring all the alternatives and deciding on a strategy will take time, and I want to make browser tests a higher priority so that people can trust stamps not to break their browser apps.

FredyC:
Ok let's review what should actually happen to have those tests running in a browser.

 * bundle stampit code with all polyfills and stuff ... done
 * bundle test code that references bundled stampit
 * execute that bundle in the browser

Back in days I used tool [Test'em](https://github.com/testem/testem) which is essentially test framework agnostic framework which runs tests in browser natively. If we give that bundle from second step, it should work quite ok.

ericelliott:
> bundle test code that references bundled stampit

This doesn't have to be a separate step from #1. Just use the test `index` as the entry point for the test bundle instead of the stampit index, and Rollup will take care of the rest.

Last I looked, Test'em was local-only, and did not run on Sauce Labs for CI. Has this changed? I was looking at [sauce-tap-runner](https://github.com/conradz/sauce-tap-runner). There's also Zuul, but it looked like a bunch of work manually hooking up the API.

FredyC:
Ok I see what you me although for ava there is no actual `index.js` because of the parallel running, but that can be added simply for sake of this. 

There is this example for testem ... https://github.com/testem/testem/tree/master/examples/saucelabs so apparently it should be possible.

ericelliott:
Yeah... if you want to work on the Ava stuff, a lot of work needs to be done in the Ava issue I linked earlier. Until Ava releases a browser-compat version, our tests will stick with Tape. Further browser test discussions should happen in https://github.com/stampit-org/stampit/issues/68

Thanks! =)

ericelliott:
BTW, feel free to re-open this issue when Ava supports browsers. Ava seems legit, and I've been wishing I could use it in a real project -- sadly... all my current work is universal JS, which means I need working browser tests. );

FredyC:
Ok I will certainly try this, but the #209 should be merged first anyway to have rollup at hand.

PR #211, opened 2016-06-01
Export isComposable and isStamp separately

FredyC:
Based on [#209](https://github.com/stampit-org/stampit/pull/209#issuecomment-222907611)

I am not much of a fan of having source file outside of src folder, but I guess this is best way for greater good.

/cc @ericelliott @koresar

### DIFF
--- a/isComposable.js
+++ b/isComposable.js
@@ -1,0 +1,5 @@
+import isObject from 'lodash/isObject';
+
+export default function isComposable(obj) {
+  return isObject(obj);
+}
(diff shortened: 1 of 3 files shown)


ericelliott:
LGTM.

koresar:
@FredyC Do you know that you can create branches in `stampit-org` repo? No need to implement things in your forks. You're a collaborator. :)

FredyC:
I know, but I am more used to forks and I have set of git aliases for that. It seems bit cleaner way to me too.

koresar:
It's just more complex for the rest of the people to collaborate. I.e. we can't fix up your branch on a fly. The conflict resolution becomes a pain in the neck too. Also, it's longer to pull the fork and test it then a branch.
But, as long as you prefer forks we all can leave with that for sure.

PR #213, opened 2016-06-02
Add a note to README that v3 is upcoming

koresar:


FredyC:
Do we need to fix those tests failing here? Last merge was already failing, so I suppose it's not relevant.

koresar:
I believe it will be fixed in v3_0 branch. No source code changes are expected to the stampit@2, so let those tests fail. We'd need to fix them though if we are going to cut a new stampit@2 release.

PR #215, opened 2016-06-13
Fix create() when passing arguments

FredyC:
Calling `this(slice(arguments))` isn't definitely same thing as using spread `this(...args)` as it will pass options wrapped in array and breaking havoc everywhere. It's lucky I had tests in my project to see this broken. I've added test for this here.

@koresar Why did you actually removed spread from `create(...args)`? Are you aware that you wont save a single byte on this? In this case Babel would transpile it to exactly same code like I've pushed :)

I've also changed little bit how tests are executed because previous form wasn't compatible on Windows. It's bit less clear that tests are run through babel too this way, but I am not sure how else to do it in compatible way except perhaps [babel-tape-runner](https://github.com/wavded/babel-tape-runner).


koresar:
Well spotted! I thought we have tests for the `create`. Sorry.

Removing `...` saved about 1K of the minified stampit.

I'm surprised it works on Windows. I wonder though why this line doesn't work well on Windows?
`"test": "babel-node --presets es2015 ./node_modules/.bin/tape test/index.js",`

koresar:
One more thing. The `...` syntax is slower than the `.apply(this, arguments)` because each `...` is transpiled to a `for` loop.

koresar:
Thank you for the fix. Will publish new RC in a moment.

koresar:
Please, try the latest `v3_0` branch on Windows. Does it work? If not, I'm happy to help fixing it.

FredyC:
> One more thing. The ... syntax is slower than the .apply(this, arguments) because each ... is transpiled to a for loop.

This simply isn't true, Babel isn't that stupid. It has seen that you are just passing all arguments to function call and it did exactly same thing `this(undefined, arguments)`. On this particular call you haven't saved anything, just made it less readable.

koresar:
You are correct. This code is less readable. This is the price we pay for performance and bundle size.

Unfortunately babel is somewhat stupid.

This ES6
```js
  methods(...args) {
    return (this.compose || compose).call(this, {methods: assign({}, ...args)});
  },
```
gets transpiled to:
```js
    methods: function methods() {
      for (var _len = arguments.length, args = Array(_len), _key = 0; _key < _len; _key++) {
        args[_key] = arguments[_key];
      }

      return (this.compose || compose).call(this, { methods: assign.apply(undefined, [{}].concat(args)) });
    },
```
Check yourself.

Stampit has about a dozen functions which accept `...args`. So, we are avoiding this repetitive `for` loop a dozen times.

koresar:
I could be wrong about the performance thing though. But I'm definitely correct about the bundle size.

FredyC:
Yea I think that `slice` is definitely much slower than using `for` loop, I remember I was investigating these things in past quite heavily. Perhaps it is different now but I wouldn't bet much on it. 

Frankly it seems ridiculous trying to save bytes on size in times of fast internet. It is surely difference to have 60K vs 5K, but having 5K or 6K is really a minor difference that nobody will care about.

koresar:
We can reimplement `slice` ourselves because it is in a separate file now (`src/slice.js`).

> Frankly it seems ridiculous trying to save bytes on size in times of fast internet.

That is true. I'd rather ask more opinions on what people would prefer: more readable stampit source code, or smaller bundle size.

FredyC:
Let's do optimizations properly based on benchmarks, no point trying to make it faster when we don't know how fast it actually is. I just recall from old days that there are some `arguments` related optimizations delivered by engine and that it actually runs slower if you fiddle with `arguments` incorrectly. Even passing them somewhere else makes that function deoptimized. I would like to believe it's different these days, but seeing that Babel does loop instead of slice, I am not convinced about it.

FredyC:
There is also very nice library called [fast.js](https://github.com/codemix/fast.js/tree/master) which focuses on reimplementing native functions in most performant way. There is nothing for treating `arguments`, but it's surely very interesting show case.

ericelliott:
`..rest` vs `.slice()` is bikeshedding and premature optimization at this point. Flip a coin and move on. ;)

ericelliott:
And by flip a coin, I mean if it ain't broke, don't fix it.

FredyC:
Yea I agree. As @koresar likes to say "premature optimization is way to hell" (or something like that) and then he goes and does premature optimization to save couple of bytes :) I guess it doesn't matter now since it's already done like this, but it was completely unnecessary for sure.

koresar:
Hey, don't change my words please. Quoting myself:
> I could be wrong about the performance thing though. But I'm definitely correct about the bundle size.

And
> Removing `...` saved about 1K of the minified stampit.

FredyC:
>  It is surely difference to have 60K vs 5K, but having 5K or 6K is really a minor difference that nobody will care about.

Yea well, I still don't agree that 1K matter really ;) But as I said, it's done so let's leave it. I am sure there will be couple more rewrites anyway once we start focusing on performance.

PR #230, opened 2016-09-12
Update advanced_examples.md

eladchen:
The docs can be misleading for people who use this example. 

Arrow functions do not expose an 'arguments' object, the only reason the example works is due to babel.

A better example is to use the correct es6 syntax, by using [rest parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters).


FredyC:
Good catch, thanks!

koresar:
@eladchen great and sneaky one. Thanks mate

eladchen:
@FredyC @koresar My pleasure.

Issue #224, opened 2016-07-27
Drop support for node.js 0.x and IE11


koresar:
How about dropping support for these two outdated platforms? :) Basically, because I don't want to introduce any new package dependencies to `stampit`.

node v0.x and IE11 do not have `Object.assign` implemented.

Many tools and modules already dropped support for node 0.x. We won't be the first.

For IE11 compatibility we would need to announce that [a polyfil](https://github.com/sindresorhus/object-assign) is required. But that's basically it.

Thoughts?

FredyC:
Well IE11 is still sort of used as not many user has migrated to Windows 10 yet, but I don't have any actual stats of how many users actually use it on Win7.

I do like more this polyfill: https://cdn.polyfill.io/v2/docs/features/ as it provides automatic detection of missing features and serves only those, but obviously it cannot be used in NodeJS environment so your recomendation is fine as well.

koresar:
If no other objections arise in couple of days we would go this direction.

koresar:
Ok. 3 days. No objections.
It's decided!
Stampit v3 is dropping support for environments which do not have `Object.assign` function or polyfil.

PR #235, opened 2016-09-22
Infrastructure update.

koresar:
Use latest ESLint, Babel, Rollup.
Rename distr. bundle files to avoid common confusions.


koresar:
Oops! ESLint v3 dropped support of node 0.12.
Should we also drop support of node 0.12?

FredyC:
I thought we already agreed on dropping support for 0.12. You even made issue for it :) https://github.com/stampit-org/stampit/issues/224

koresar:
I admire your memory! Dropping it now (because I dropped only the 0.10). :)

FredyC:
It took me a while to find it, I only vaguely knew, we talked about it.

.eslintrc
```diff
@@ -1,12 +1,13 @@
...
+  "extends": "airbnb-base",
   "rules": {
     "object-curly-spacing": [2, "never"],
     "no-param-reassign": 0,
     "comma-dangle": 0,
     "no-confusing-arrow": 0,
     "prefer-rest-params": 0,
-    "no-use-before-define": 1
+    "no-use-before-define": 1,
+    "no-plusplus": 0
```
FredyC: Scratch that only occurrence, it's also in some for loops. But still I don't like that operator :)

PR #236, opened 2016-09-22
Fix some lint issues and update airbnb

FredyC:
I've replaced for loops using `++` operator with `reduce` which I would say is better in general.

Change for a rule "no-use-before-define" allows hoisting of functions which is necessary to pass with green lights :)

Updated airbnb since it was complaining about new version of eslint. Nothing for broken because of that.

### DIFF
--- a/src/merge.js
+++ b/src/merge.js
@@ -23,7 +23,7 @@ function mergeOne(dst, src) {
   const returnValue = isObject(dst) ? dst : {};
 
   const keys = Object.keys(src);
-  for (let i = 0; i < keys.length; i++) {
+  for (let i = 0; i < keys.length; i += 1) {
     const key = keys[i];
 
     const srcValue = src[key];
--- a/.eslintrc
+++ b/.eslintrc
@@ -7,7 +7,6 @@
     "comma-dangle": 0,
     "no-confusing-arrow": 0,
     "prefer-rest-params": 0,
-    "no-use-before-define": 1,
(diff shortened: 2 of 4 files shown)


src/merge.js
```diff
@@ -22,23 +22,19 @@ function mergeOne(dst, src) {
   // See if 'dst' is allowed to be mutated. If not - it's overridden with a new plain object.
   const returnValue = isObject(dst) ? dst : {};
 
-  const keys = Object.keys(src);
-  for (let i = 0; i < keys.length; i++) {
-    const key = keys[i];
-
+  return Object.keys(src).reduce((result, key) => {
```
koresar: This `for` loop was a performance optimization.

The problem with reduce is the performance. We are creating one more object (function) and scope. Accessing variables of the upper scope is slower than accessing from own scope.
More info here: https://github.com/petkaantonov/bluebird/wiki/Optimization-killers

Please, revert the loop (and the other one). Stampit has to be fast!

koresar: Sorry. That's the right link: https://reaktor.com/blog/javascript-performance-fundamentals-make-bluebird-fast/

FredyC: Sorry, but performance is relative till you introduce actual performance tests. I would prefer clarity of code till we know this is a bottleneck...

abhisekp: Isn't that statement application for apps and not libraries? :smirk:

Issue #246, opened 2016-10-01
Direct import of stampit functions is broken with rollup.


FredyC:
Well this is pretty nasty and got my head spinning for a bit. I am using stampit within my library which is then bundled with rollup. In the source code I have:

```js
import { init } from 'stampit';
init(...).methods(...)
```

Rollup makes this out of it:
```js
var stampit = require('stampit');
stampit.init(...).methods(...)
```

Because rollup made it like a call on object of the module instead of a direct function, the pure compose is suddenly used as it's present on `stampit.compose`. Then obviously chaining gets broken because pure compose is missing those functions.

```js
export function initializers(...args) {
  return ((this && this.compose) || stampit).call(this, {
    initializers: extractFunctions(...args)
  });
}
```

I am fairly certain that stampit shouldn't export pure compose at all. Since it's stampit export, it shouldn't provide pure compose. For anyone who wants that, just `import compose from 'stampit/compose` is way more easier to comprehend anyway.

koresar:
Thanks! Fixed in v3.0.2

FredyC:
I confirm it's working now. Great work, @koresar

PR #247, opened 2016-10-01
[WIP] Broken chaining with direct module

FredyC:
Not for merge yet, just to show failing tests demonstrating #246


FredyC:
Huh, that's strange. Test appears to be passing here, but in log there is test fail clearly.. https://travis-ci.org/stampit-org/stampit/jobs/164304698

koresar:
Looks like two bugs collided and annihilated. @FredyC thanks for the PR with the proper test.

I pushed two more commits:
1) Fix the Travis exit code (removed the `tap-dot` completely)
2) Export the right `compose` function as we designed

PR #249, opened 2016-10-02
Remove "eslint-plugin-import" and "babel-eslint" dependencies as not needed any more

koresar:


FredyC:
I don't think you can remove the import plugin, it's required by airbnb. Also this statement of yours isn't correct:

> AirBNB eslint configuration already lints import/exports for us.

That's the plugin which does linting of import/exports, nothing else.

koresar:
Okay. Another way. 
This page https://www.npmjs.com/package/eslint-config-airbnb-base says to run the little shell [script](https://www.npmjs.com/package/eslint-config-airbnb-base#eslint-config-airbnb-base-1).
It upgraded eslint, set the new version for the eslint-config-airbnb-base and eslint-plugin-import. See the last commit.

FredyC:
Yes this should be alright. I run into same problem with my library. I suppose that airbnb will get updated soon to support import plugin 2.x. Greenkeeper will let us know.

PR #251, opened 2016-10-10
Warning about Object.assign polyfill

koresar:
UNPKG would list the available distributable files instead of the node.js/CJS one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Stampit uses [three different kinds of prototypal OO](https://vimeo.com/69255635
 [![NPM](https://nodei.co/npm/stampit.png)](https://www.npmjs.com/package/stampit)
 
 Browsers: [![CDNJS](https://img.shields.io/badge/cdnjs.com--green.svg)](https://cdnjs.com/libraries/stampit)
-[![UNPKG](https://img.shields.io/badge/unpkg.com--green.svg)](https://unpkg.com/stampit@latest/dist/)
+[![UNPKG](https://img.shields.io/badge/unpkg.com--green.svg)](https://unpkg.com/stampit)
 
 **WARNING!** If running in node.js <= v0.12 or IE <= 11 then you'd need to polyfill the `Object.assign`. Like [this](https://github.com/sindresorhus/object-assign), or [that](https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Object/assign/polyfill.js) ([autodetected](https://cdn.polyfill.io/v2/polyfill.min.js?features=Object.assign)).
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -49,7 +49,9 @@ Stampit uses [three different kinds of prototypal OO](https://vimeo.com/69255635
 [![NPM](https://nodei.co/npm/stampit.png)](https://www.npmjs.com/package/stampit)
 
 Browsers: [![CDNJS](https://img.shields.io/badge/cdnjs.com--green.svg)](https://cdnjs.com/libraries/stampit)
-[![UNPKG](https://img.shields.io/badge/unpkg.com--green.svg)](https://unpkg.com/stampit)
+[![UNPKG](https://img.shields.io/badge/unpkg.com--green.svg)](https://unpkg.com/stampit@latest/dist/)
+
+**WARNING!** If running in node.js <= v0.12 or IE <= 11 then you'd need to polyfill the `Object.assign`. Like [this](https://github.com/sindresorhus/object-assign), or [that](https://github.com/Financial-Times/polyfill-service/blob/master/polyfills/Object/assign/polyfill.js), or [else](https://npms.io/search?q=keywords%3Aponyfill).
```
FredyC: I would direct the polyfill service directly to actual file that can be served: https://cdn.polyfill.io/v2/polyfill.min.js?features=Object.assign. It's not very easy to find information how to make such request. However if you open that in chrome, you will get (almost) empty file as polyfill is not needed there, so that might be misleading bit.

Also I don't think it's relevant to use ponyfill link there. It won't help consumer because ponyfill is just module that is used directly in the code, it's not used globally. And for Object.assign it will lead you to already mentioned sindresorhus's module anyway.

koresar: I'm lost. Could you please provide what you wanna see here? I don't understand. :)
Just push to this branch. Or edit on the GitHub. Whatever is easier for you. :)
Thanks!

README.md
```diff
@@ -49,7 +49,9 @@ Stampit uses [three different kinds of prototypal OO](https://vimeo.com/69255635
 [![NPM](https://nodei.co/npm/stampit.png)](https://www.npmjs.com/package/stampit)
 
 Browsers: [![CDNJS](https://img.shields.io/badge/cdnjs.com--green.svg)](https://cdnjs.com/libraries/stampit)
-[![UNPKG](https://img.shields.io/badge/unpkg.com--green.svg)](https://unpkg.com/stampit)
+[![UNPKG](https://img.shields.io/badge/unpkg.com--green.svg)](https://unpkg.com/stampit@latest/dist/)
```
FredyC: I don't think this is a good way to show everything as it may confuse people what to use actually. CDNJS is also providing UMD build only.

And there is no point in serving anything else but UMD bundle for a browser. Those other two are using `require` which would not work without bundler and in that case people can probably use NPM directly as well.

FredyC: It should stay as it was ;)

FredyC:
@koresar I've renamed the PR as change to unpkg is gone :) Better to change commit message when squashing as well.

PR #253, opened 2016-10-10
Fix release links in readme

FredyC:
Thanks @abhisekp for reporting this


koresar:
Thanks for fixing the second link! 
But the first link must be left intact please 
https://github.com/stampit-org/stampit/releases/tag/2.0

Those breaking changes links. This is important.

FredyC:
Interesting, I don't know why I thought that both links are broken :) Good catch, it's reverted

PR #254, opened 2016-10-10
perf(extract-functions): Refactor for better performance

abhisekp:
This will result in much faster flattening of arrays and function extraction to a single array.

**Demo (babel repl)**: http://tinyurl.com/extract-fns

### DIFF
--- a/src/extract-functions.js
+++ b/src/extract-functions.js
@@ -1,11 +1,17 @@
 import isFunction from './isFunction';
 
-export default function (...args) {
-  let result = [];
-  for (let i = 0; i < args.length; i += 1) {
-    const arg = args[i];
-    if (isFunction(arg)) result.push(arg);
-    else if (Array.isArray(arg)) result = result.concat(arg.filter(isFunction));
+export default function extractFunctions() {
+  const fns = [];
+
+  for (let i = 0; i < arguments.length; i += 1) {
+    const arg = arguments[i];
+
+    if (isFunction(arg)) {
+      fns.push(arg);
+    } else if (Array.isArray(arg)) {
+      fns.push(...(extractFunctions(...arg) || []));
+    }
   }
-  return result.length === 0 ? undefined : result;
+
+  return fns.length === 0 ? undefined : fns;
 }


koresar:
Awesome! I'm going  measure the performance of this and merge sometime today. 

Thanks!

abhisekp:
# TODO
Explain the algorithm in comments

src/extract-functions.js
```diff
@@ -1,11 +1,17 @@
...
+export default function extractFunctions() {
+  const fns = [];
+
+  for (let i = 0; i < arguments.length; i += 1) {
+    const arg = arguments[i];
+
+    if (isFunction(arg)) {
+      fns.push(arg);
+    } else if (Array.isArray(arg)) {
+      fns.push(...(extractFunctions(...arg) || []));
```
FredyC: Perhaps instead of pushing empty array it might be better to actually check its length before doing so. It's probably micro optimization and could use some measurement. Also it might be more readable. I had to read this line 3 times to actually understand what's happening there.

src/extract-functions.js
```diff
@@ -1,11 +1,17 @@
 import isFunction from './isFunction';
 
-export default function (...args) {
-  let result = [];
-  for (let i = 0; i < args.length; i += 1) {
-    const arg = args[i];
-    if (isFunction(arg)) result.push(arg);
-    else if (Array.isArray(arg)) result = result.concat(arg.filter(isFunction));
+export default function extractFunctions() {
+  const fns = [];
```
FredyC: Why to rename to `fns`? I would say that `result` is much better to mark something that will be returned.

src/extract-functions.js
```diff
@@ -1,11 +1,17 @@
...
+
+    if (isFunction(arg)) {
+      fns.push(arg);
+    } else if (Array.isArray(arg)) {
+      fns.push(...(extractFunctions(...arg) || []));
+    }
   }
-  return result.length === 0 ? undefined : result;
+
+  return fns.length === 0 ? undefined : fns;
```
FredyC: I am aware that this was there before, but I am actually wondering why we are returning `undefined` here when no functions are found. It's not against spec and returning empty array will simply cause no iteration. 

It would also make the above recursive call better in the sense it wont need to create empty array knowing that this function always returns array.

@koresar ?

koresar: If you remove the `undefined` return value then the tests would break. :)
This function must return undefined in no functions found in its arguments. So, yeah, `undefined` is necessary.

koresar:
My perf measurements.

Run them from here: https://github.com/koresar/bench-node
These tests run the transpiled ES5 code, not the ES6 shown above.

FLATTEN 1 - the previous implementation (.min.gz file size 1.59KB)
FLATTEN 2 - the implementation from this PR (.min.gz file size 1.62KB).
LODASH FF - is this: `_.filter(_.flatten(args), isFunction`

* If there is no recursion then both variants are similar.
```
> node flatten/2
v4.6.0 1 item with undefined
FLATTEN 1   107660.1  | malloc: >= 14.69  MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
FLATTEN 2   110704.73 | malloc: >= 16.47  MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
LODASH FF   114341.72 | malloc: >= 14.73  MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
Fastest is LODASH FF
```

* If case the incoming arguments are complex, like array of arrays, then both implementations are identical in CPU, but the second one takes slightly more memory.
```
> node flatten/1
v4.6.0 10 items
FLATTEN 1   87161.47  | malloc: >= 15.67  MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
FLATTEN 2   87590.96  | malloc: >= 30.49  MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
LODASH FF   100668.87 | malloc: >= 14.73  MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
Fastest is LODASH FF
```

P.S. @abhisekp, you probably wanna apply the "loose" changes to this PR to minimize the output bundle. It might even speed things up. Feel free to run the test above yourself. :)

koresar:
@abhisekp proposed a third implementation of the function.
```js
import isFunction from './isFunction';

const concat = Array.prototype.concat;
export default function () {
  const fns = concat.apply([], arguments).filter(isFunction);
  return fns.length === 0 ? undefined : fns;
}
```

Here are the perf. results across all node.js versions on my laptop.

10 items array.
```
> ./all_nodes.sh flatten/1
v0.10.47 10 items
FLATTEN 1   65773.45  | malloc: >= 28.74  MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
FLATTEN 2   65356.04  | malloc: >= 28.52  MB | GC calls: >= 2   | Mem/cycle: >= ~ 0       
FLATTEN 3   64799.63  | malloc: >= 21.64  MB | GC calls: >= 2   | Mem/cycle: >= ~ 0       
LODASH FF   69509.9   | malloc: >= 6.84   MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
Fastest is LODASH FF

v0.12.16 10 items
FLATTEN 1   48955.46  | malloc: >= 29.71  MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
FLATTEN 2   47523.45  | malloc: >= 12.72  MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
FLATTEN 3   48458.25  | malloc: >= 1.97   MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
LODASH FF   53024.4   | malloc: >= 2.95   MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
Fastest is LODASH FF

v4.6.0 10 items
FLATTEN 1   88213.83  | malloc: >= 16.65  MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
FLATTEN 2   92537.16  | malloc: >= 16.72  MB | GC calls: >= 1   | Mem/cycle: >= ~ 0       
FLATTEN 3   95071.15  | malloc: >= 6.89   MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
LODASH FF   104188.03 | malloc: >= 5.87   MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
Fastest is LODASH FF

v5.12.0 10 items
FLATTEN 1   87573.79  | malloc: >= 16.65  MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
FLATTEN 2   96284.76  | malloc: >= 18.68  MB | GC calls: >= 1   | Mem/cycle: >= ~ 0       
FLATTEN 3   96540.67  | malloc: >= 5.91   MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
LODASH FF   104437.3  | malloc: >= 5.87   MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
Fastest is LODASH FF

v6.7.0 10 items
FLATTEN 1   62318.56  | malloc: >= 15.00  MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
FLATTEN 2   65788.09  | malloc: >= 22.00  MB | GC calls: >= 1   | Mem/cycle: >= ~ 0       
FLATTEN 3   68186.61  | malloc: >= 6.00   MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
LODASH FF   75645.36  | malloc: >= 6.00   MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
Fastest is LODASH FF
```

2 items array
```
> ./all_nodes.sh flatten/2
v0.10.47 1 item with undefined
FLATTEN 1   76882.66  | malloc: >= 12.75  MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
FLATTEN 2   62167.34  | malloc: >= 38.61  MB | GC calls: >= 2   | Mem/cycle: >= ~ 0       
FLATTEN 3   62201.21  | malloc: >= 16.73  MB | GC calls: >= 2   | Mem/cycle: >= ~ 0       
LODASH FF   74590.92  | malloc: >= 4.87   MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
Fastest is FLATTEN 1

v0.12.16 1 item with undefined
FLATTEN 1   52772.96  | malloc: >= 30.69  MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
FLATTEN 2   49348.26  | malloc: >= 11.75  MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
FLATTEN 3   48268.43  | malloc: >= ~ 0       | GC calls: >= 0   | Mem/cycle: >= ~ 0       
LODASH FF   53110.37  | malloc: >= 1.97   MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
Fastest is LODASH FF,FLATTEN 1

v4.6.0 1 item with undefined
FLATTEN 1   113933.3  | malloc: >= 14.68  MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
FLATTEN 2   87618.49  | malloc: >= 22.59  MB | GC calls: >= 1   | Mem/cycle: >= ~ 0       
FLATTEN 3   86366.92  | malloc: >= ~ 0       | GC calls: >= 0   | Mem/cycle: >= ~ 0       
LODASH FF   108488.16 | malloc: >= 5.87   MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
Fastest is FLATTEN 1

v5.12.0 1 item with undefined
FLATTEN 1   107333.9  | malloc: >= 16.65  MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
FLATTEN 2   88726.87  | malloc: >= 21.62  MB | GC calls: >= 1   | Mem/cycle: >= ~ 0       
FLATTEN 3   87040.05  | malloc: >= 4.89   MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
LODASH FF   108251    | malloc: >= 5.87   MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
Fastest is LODASH FF,FLATTEN 1

v6.7.0 1 item with undefined
FLATTEN 1   88129.24  | malloc: >= 23.00  MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
FLATTEN 2   66571.43  | malloc: >= 18.00  MB | GC calls: >= 1   | Mem/cycle: >= ~ 0       
FLATTEN 3   68384.86  | malloc: >= 6.02   MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
LODASH FF   84992.47  | malloc: >= 7.00   MB | GC calls: >= 0   | Mem/cycle: >= ~ 0       
Fastest is FLATTEN 1
```

P.S. The file size:
Current stampit impl.: `Estimating dist/stampit.umd.min.js: 4.58 KB, GZIP : 1.59 KB`
The code from above: `Estimating dist/stampit.umd.min.js: 4.5 KB, GZIP : 1.56 KB`

The new code uses much less memory than any other implementation. The perf is about the same to some extent. The file size is a bit smaller. Code is more readable. I like it in general. :)

Although, I'd rather avoid discussions about micro benchmarking.

FredyC:
I kinda think that we are hunting performance in wrong places here. In practice I had no need so far to use a complex structure when specifying initializers. I don't see use case why would anyone pass array of arrays down there... Perhaps it might be better to optimize by actually excluding whole flattening completely. I know, it's breaking change and should have come before V3 probably, but maybe we can think of it for V4 in a future.

koresar:
Yeah. I thought of removing flattening sometime next major release. We need that flattening at least for [this](https://github.com/stampit-org/stampit/blob/master/src/standardise-descriptor.js#L57) line to work.

Surely, none passes arrays of arrays as initializers. It won't even work in the current code base. However, the "2 items array" test is the real world test. That's exactly how the `extractFunctions` is called most of the times.

src/extract-functions.js
koresar: That's easy. Just make the change yourself and run. I'm very busy atm. Can't really spend time explaining.

FredyC:
Yea I agree, but extracting only functions from flat array is as simple as `initializers.filter(isFunction)`, no need for extra file with additional code for that.

koresar:
Okay. @FredyC has some bright ideas. So, here what we'll do.
* I'll apply the proposed code because of the single reason - it's shorter.
* Than, if time allows, @FredyC will supply a new PR brining all the great ideas together.

I'll close this PR and apply @abhisekp's code separately.

PR #266, opened 2016-10-18
Remove "babel" field in package.json for react-native support

Piroro-hs:
React-native-packager try to use babel config in package.json or .babelrc and throw "Unknown plugin" error.
This PR will solve this issue and allow using stampit with react-native.

https://github.com/facebook/react-native/issues/4062
[react-native TransformError when requiring a node_module that has a package.json with a babel config that specifies uninstalled plugin (stackoverflow)](http://stackoverflow.com/questions/37521478/react-native-transformerror-when-requiring-a-node-module-that-has-a-package-json)

### DIFF
--- a/build/index.js
+++ b/build/index.js
@@ -13,7 +13,9 @@ import MagicString from 'magic-string';
 // That plugin made the code too large and also is hard to configure.
 // The main idea is to make sure babel does not transpile ES6 modules to CJS
 // because Rollup handles import/export itself.
-const es2015Plugins = pkg.babel.plugins;
+const es2015Presets = [
+  ['es2015', {loose: true, modules: false}]
+];
 
 const moduleName = 'stampit';
 
@@ -25,7 +27,7 @@ function execute() {
       {
         format: 'cjs',
         ext: '.full.js',
-        babelPlugins: es2015Plugins,
+        babelPresets: es2015Presets,
         dest: 'dist',
         moduleName: 'stampit'
       }
@@ -36,7 +38,7 @@ function execute() {
       {
         format: 'umd',
         ext: '.umd.js',
-        babelPlugins: es2015Plugins,
+        babelPresets: es2015Presets,
         dest: 'dist',
         moduleName: 'stampit'
       }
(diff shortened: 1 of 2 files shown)


koresar:
Wow. Awesome! 
We had to put all that code for a reason probably not relevant anymore. 
Let me test it back and forth. There might be caveats. 
Will publish ASAP. 

Cheers

FredyC:
Looks good to me as well. I remember (although bit foggy) that we had to do it like this because of ES2015 preset transpiling those ES modules and babel-preset-es2015-rollup wasn't working that well.

It's bit problematic, that build might actually pass, but don't have exactly tools to automatically verify that those builds are really working properly.

koresar:
First of all, thank you very much @Piroro-hs for that cool deep techy contribution. 

I checked the final resulting bundle.

1) The new bundle now has this code:
```js
var _typeof = typeof Symbol === "function" && typeof Symbol.iterator === "symbol" ? function (obj) { return typeof obj; } : function (obj) { return obj && typeof Symbol === "function" && obj.constructor === Symbol && obj !== Symbol.prototype ? "symbol" : typeof obj; };
```
which replaces the `typeof` used in the `isObject()` and `isPlainObject()`. It's done to support symbols I assume.
AFAIK we don't need Symbols support **in these two methods**.

2) The second issue is that the code above is repeated twice, which increases bundle size for no gain.

Is there a way to get rid of that "symbol" support in those functions?

=====

P.S.
Again, this happened again! Bloody ES6, bloody babel, bloody transpiling! I wanna my ES5 days back!
I'm certain that if the code was ES5 from the beginning we won't ever need PR's like that. :)

FredyC:
Babel is obviously over complicated, let's check out Bublé :) #267

koresar:
It would be great is `React-native-packager` would behave same as node.js - use the `main` file from package.json.

I'm liking the Buble approach very much. @Piroro-hs would #267 resolve your issue?

Piroro-hs:
I've just cloned the https://github.com/stampit-org/stampit/pull/267 repo and it works perfectly!(^^)

koresar:
Great! Closing this one. :)

koresar:
@Piroro-hs published stampit v3.0.4
😃

PR #267, opened 2016-10-19
[WIP] Use Bublé instead of cursed Babel

FredyC:
Just little experiment to contradict #266 with this little [unknown transpiler](buble.surge.sh). First impression is that it's faster then babel and there is couple of bytes saved as well. I absolutely love that it doesn't need **any** configuration. Just use plugin and done.

**Current babel configuration (master)**
Estimating dist/stampit.umd.min.js: 4.5 KB, GZIP : 1.56 KB

**Using Babel 2015 preset (@Piroro-hs)**
Estimating dist/stampit.umd.min.js: 4.96 KB, GZIP : 1.63 KB

**Using Buble**
Estimating dist/stampit.umd.min.js: 4.5 KB, GZIP : 1.52 KB

Here I made a package of all three approaches for comparison.
[stampit-dist.zip](https://github.com/stampit-org/stampit/files/538068/stampit-dist.zip)

The PR needs some cleanup, just wanted to do a quick check.


koresar:
Wow. Cool! Can we kill `package.json->babel` too? Please :)

FredyC:
Ok, got rid of Babel, although we still need `babel-cli` to actually run the build because Buble apparently doesn't have alternative to `babel-node` :(

koresar:
Looks like this PR needs to be improved because the folder still contains the bloody `.babelrc` file. That's one of the issues with the react-native.

FredyC:
Yea right, I've missed that :)

PR #270, opened 2016-10-29
Obj create perf fix

koresar:
* Fix property access perf.
* Fix some JSDoc comments.
* Run CI tests for node v7 too.
* Add benchmarking test, make it run in CI only.

### DIFF
--- a/src/compose.js
+++ b/src/compose.js
@@ -32,7 +32,7 @@ function createFactory(descriptor) {
 
 /**
  * Returns a new stamp given a descriptor and a compose function implementation.
- * @param {object} [descriptor={}] The information about the object the stamp will be creating.
+ * @param {Descriptor} [descriptor={}] The information about the object the stamp will be creating.
  * @param {Compose} composeFunction The "compose" function implementation.
  * @returns {Stamp}
  */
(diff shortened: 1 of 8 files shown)


src/compose.js
```diff
@@ -7,12 +7,13 @@ const assign = Object.assign;
...
  * Creates new factory instance.
- * @param {object} descriptor The information about the object the factory will be creating.
+ * @param {Descriptor} descriptor The information about the object the factory will be creating.
  * @returns {Function} The new factory function.
  */
 function createFactory(descriptor) {
   return function Stamp(options, ...args) {
-    const obj = Object.create(descriptor.methods || {});
+    // Next line was optimized for most JS VMs. Please, be careful here!
+    const obj = Object.create(descriptor.methods || null);
```
FredyC: I am wondering, have you actually tried to avoid using `Object.create` itself if methods are not set? You could do something like `descriptors.methods ? Object.create(descriptor.methods) : {};`. I do remember from some benchmarks I was running like 2-3 years ago that this function itself was terribly slow. Also I don't actually see amy point in executing it if you just want a new object

FredyC: Using `new` would mean something like this if I am not mistaken, but I would be really surprised if it would be really faster.
```js
function createFactory(descriptor) {
  function Stamp(options, ...args) {
    merge(this, descriptor.deepProperties);
    assign(this, descriptor.properties);
    Object.defineProperties(this, descriptor.propertyDescriptors || {});
    ...
  }
  if (descriptor.methods) {
    Stamp.prototype = descriptor.methods;
  }
  return (...args) => new Stamp(...args)
}
```

ericelliott: @FredyC - Because of V8's hidden classes optimization, using new can be up to **10 times faster** for property access (not instantiation) than `Object.create()`.

Of course we should profile both instantiation and property access. We also need to run the check and loop about a million times to see all the optimizations kick in -- on modern CPUs, that shouldn't take terribly long.

koresar:
Thanks @FredyC !

Issue #273, opened 2016-11-04
Simplest example displays the wrong output.


kurtpeters:
I noticed the last two commented outputs from the "simplest example" are incorrect. Since the initializer is destructuring the argument
```
const MyStamp = stampit()       // create new empty stamp
.props({                        // add properties to your future objects
  myProp: 'default value'
})
.methods({                      // add methods to your future objects
  getMyProp() {
    return this.myProp;
  }
})
.init(function ({value}) {      // add initializers to your future objects
  this.myProp = value || this.myProp;
})
```
should the following
```
console.log(MyStamp('new value'));             // { myProp: 'new value' }
console.log(MyStamp('new value').getMyProp()); // new value
```
be documented to reflect to the destructuring from the initializer?
```
console.log(MyStamp({value: 'new value'}));             // { myProp: 'new value' }
console.log(MyStamp({value: 'new value'}).getMyProp()); // new value
```

FredyC:
Good catch, thanks! Would you mind creating PR?

kurtpeters:
No problem! Here's the PR - https://github.com/stampit-org/stampit/pull/274

PR #274, opened 2016-11-04
fix(readme): Properly document destructuring usage in simplest example

kurtpeters:
https://github.com/stampit-org/stampit/issues/273


FredyC:
Thanks!

PR #277, opened 2016-11-12
Few devDependency upgrades

koresar:
* Remove annoying dependency - eslint-config-airbnb-base
* Upgrade some devDependencies.
* The eslint-plugin-import was never used! Turned in on.


FredyC:
Annoying dependency? I don't get it. It contains many useful rules you have simply disabled with this. Also import plugin was used and configured with airbnb.

koresar:
Every time it upgrades - something breaks. :(

FredyC:
That's not true. There was this incompatibility with import plugin, but that's fixed for some now.

Also I don't understand how could there be so many outdated dependecies, I thought greenkeeper is watching it or did you disabled it?

koresar:
I did not. Greenkeeper doesn't do it's job good enough.

PR #283, opened 2016-12-08
Implement `composers` feature

koresar:
- Implement `composers` feature in a specs compatible way.
- Move all internal function to the utils.js for simplicity.
- Performance optimize object creation (create less closure objects)
- Performance optimize composition (create less closure objects)

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -105,6 +105,7 @@ The arguments can be either another stamps, or the following structure:
  * `@param  {Object} [options.properties]` Same as `options.props`
  * `@param  {Object} [options.init]` A closure (function) used to create private data and privileged methods
  * `@param  {Object} [options.initializers]` Same as `options.init`
+ * `@param  {Object} [options.composers]` (EXPERIMENTAL) Similar to initializers, but executed at the composition time
  * `@param  {Object} [options.deepProps]` An object to be deeply cloned into each newly stamped object
  * `@param  {Object} [options.deepProperties]` Same as `options.deepProps`
  * `@param  {Object} [options.statics]` A map of property names and values to be mixed onto stamp itself
@@ -129,6 +130,7 @@ Returns a new factory function (called a stamp) that will produce new objects.
  * `@return {Function} stamp.properties` Same as `stamp.props`
  * `@return {Function} stamp.init` Add an initializer which called on object instantiation. Returns a new stamp
  * `@return {Function} stamp.initializers` Add an initializer which called on object instantiation. Returns a new stamp
+ * `@return {Function} stamp.composers` (EXPERIMENTAL) Add a composer function which is called on stamp composition. Returns a new stamp
  * `@return {Function} stamp.deepProps` Add deeply cloned properties to the produced objects. Returns a new stamp
  * `@return {Function} stamp.deepProperties` Same as `stamp.deepProps`
  * `@return {Function} stamp.static` Add properties to the factory object. Returns a new stamp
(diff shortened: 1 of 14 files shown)


src/compose.js
```diff
@@ -22,11 +19,18 @@ function createFactory(descriptor) {
     if (!descriptor.initializers || descriptor.initializers.length === 0) return obj;
 
     if (options === undefined) options = {};
-    return descriptor.initializers.filter(isFunction).reduce((resultingObj, initializer) => {
-      const returnedValue = initializer.call(resultingObj, options,
-        {instance: resultingObj, stamp: Stamp, args: [options].concat(args)});
-      return returnedValue === undefined ? resultingObj : returnedValue;
-    }, obj);
+    const inits = descriptor.initializers;
```
koresar: This is a performance optimization. Removed unnecessary closure object creation.

test/stampit-shortcuts.js
```diff
@@ -25,7 +25,7 @@ test('stampit.refs shortcut', (t) => {
 });
 
 test('stampit.init shortcut', (t) => {
-  const init = {method1() {}};
+  const init = () => {};
```
koresar: Fixed a test. Stampit v3 does not support objects any more.

FredyC:
Hm, not sure why you want review from me. I am big fan of this change really :) I will try to have a look later on. You consider this merge-ready then?

koresar:
At least API.md better be reviewed.

koresar:
Will merge the PR in 24h.

docs/API.md
```diff
@@ -275,6 +277,79 @@ MyStamp().clone().clone().clone().x === 42; // true
...
+```
+
+#### Examples
+
+Collect every composable the stamp was created from.
+```js
+const ComponentsMonitoring = stampit().composers(({stamp, composables}) => {
+  const conf = stamp.compose.configuration || {};
+  conf._wasComposedOf = _.uniq(composables.concat(conf._wasComposedOf));
+  stamp.compose.configuration = conf;
```
koresar: Mutating it is. By design.
The mutation safely adds a single property. That's it.
Returning a new stamps here would be more unexpected than mutating existing. I tried.

src/stampit.js
```diff
@@ -1,17 +1,15 @@
...
 
-const assign = Object.assign;
-
 function createUtilityFunction(propName, action) {
   return function composeUtil() {
-    const descriptor = {};
-    descriptor[propName] = action({}, ...arguments);
-    return ((this && this.compose) || stampit).call(this, descriptor);
+    return ((this && this.compose) || stampit).call(this, {
+      [`${propName}`]: action({}, ...arguments)
```
koresar: Fixed. :)

docs/API.md
FredyC: Hm and what were implications of returning stamp? The stamps are kinda know for being immutable, so I am bit surprised that suddenly we would promote mutating?

FredyC: > Below is the usage of our new infected compose. I would like to remind that Stamps are immutable. Every compose call creates a new Stamp.

From your [Fun with stamps](https://medium.com/@koresar/fun-with-stamps-episode-8-tracking-and-overriding-composition-573aa85ba622#.bbn9swv1b) :)

FredyC: You are saying that you want people to experiment, but what if someone actually wants to avoid mutation with composers? It wont be possible like this. On other hand mutability is available directly. So what about just checking if something else but `undefined` is returned and use that instead of current stamp?

koresar: To avoid mutations with composers a user would need to `delete stamp.compose.deepConfiguration.composers`.

> So what about just checking if something else but undefined is returned and use that instead of current stamp?

It's actually how ti works. See `stampit.js`.

FredyC: Ok, I see. I should have probably read updated docs more closely :) In that case it's fine like this.

PR #295, opened 2017-01-07
chore(package): update rollup to version 0.41.1

FredyC:
Closes #294

https://greenkeeper.io/

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "deps:extra": "dependency-check package.json --extra --no-dev --ignore",
     "audit": "nsp check",
     "precheck": "npm test",
-    "check": "npm run audit && npm run deps && npm outdated --depth 0",
+    "check": "npm run audit && npm run deps",
     "watch": "watch 'clear && npm -s test' test/ src/"
   },
   "license": "MIT",
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "nsp": "^2.1.0",
     "require-all": "^2.0.0",
     "rimraf": "^2.3.4",
-    "rollup": "^0.38.2",
+    "rollup": "^0.41.1",
     "rollup-plugin-buble": "^0.15.0",
     "rollup-plugin-filesize": "^1.0.0",
     "rollup-plugin-uglify": "^1.0.0",
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "prepublish": "npm run check",
     "pretest": "npm run build",
     "test": "babel-node --presets=es2015 test && npm run lint",
+    "posttest": "node test/benchmark",
     "clean": "rimraf dist/*",
     "lint": "eslint src && eslint test",
     "prebuild": "npm run clean",
--- a/test/benchmark/.eslintrc
+++ b/test/benchmark/.eslintrc
@@ -1,6 +1,8 @@
 // Add our overrides as necessary
 {
   "rules": {
-    "global-require": 0
+    "global-require": 0,
+    "strict": 0,
+    "lines-around-directive": 0
   }
 }
--- a/test/benchmark/.eslintrc
+++ b/test/benchmark/.eslintrc
@@ -1,0 +1,6 @@
+// Add our overrides as necessary
+{
+  "rules": {
+    "global-require": 0
+  }
+}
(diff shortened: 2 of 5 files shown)


FredyC:
@koresar Can you eventually have a look why would that single test be failing on Node 7? My guess is that it is taking too long there for some reason...

```
# benchmarking property access
not ok 231 object instances property access must be as fast as plain object.
  ---
    operator: ok
    expected: true
    actual:   false
    at: Suite.<anonymous> (/home/travis/build/stampit-org/stampit/test/benchmark/property-access.js:71:9)
  ...
# Stampit x 11,255 ops/sec ±1.21% (70 runs sampled). Plain object x 27,200 ops/sec ±0.89% (91 runs sampled)
# stampit().compose()
```

koresar:
I shell try. Thanks for bringing the attention.

koresar:
I HATE TRASNPILATION!

If you run it lie so:
```
babel-node test/benchmark/property-access.js 
```
All good. Stampit objects are even faster than plain objects.

But if you run it like so (under node v7.4)
```
babel-node --presets=es2015 test/benchmark/property-access.js 
```
it fails!!! Stampit prop access is 10x slower.

Source code of the file is below:
<details>
<summary>Click to expand</summary>
const Benchmark = require('benchmark');
const newObj = () => ({components: {position: {}, velocity: {vx: 10, vy: 10}}});

const COUNT = 5000;
const stampitEntities = Array.from(Array(COUNT)).map(() => Object.assign(Object.create(null), newObj()));
const normalEntities = Array.from(Array(COUNT)).map(() => ({
  components: {
    position: {
      x: 10,
      y: 10,
    },
    velocity: {
      vy: 10,
      vx: 10,
    },
  },
}));

function runArray(arr) {
  const DELTA = 16;
  for (let i = 0; i < COUNT; i += 1) {
    const entity = arr[i].components;

    entity.position.x += (DELTA / 1000) * entity.velocity.vx;
    entity.position.y += (DELTA / 1000) * entity.velocity.vy;
  }
}

const suite = new Benchmark.Suite();
const results = [];
suite
  .add('Stampit', () => {
    runArray(stampitEntities);
  })
  .add('Plain object', () => {
    runArray(normalEntities);
  })
  .on('cycle', (event) => {
    results.push(String(event.target));
  })
  .on('complete', function () {
    console.log(results.join('. '));
  })
  .run();
</details>

package.json
```diff
@@ -54,6 +54,7 @@
     "prepublish": "npm run check",
     "pretest": "npm run build",
     "test": "babel-node --presets=es2015 test && npm run lint",
+    "posttest": "node test/benchmark",
```
koresar: Need to run benchmark tests without babel-node.

package.json
```diff
@@ -63,7 +64,7 @@
     "deps:extra": "dependency-check package.json --extra --no-dev --ignore",
     "audit": "nsp check",
     "precheck": "npm test",
-    "check": "npm run audit && npm run deps && npm outdated --depth 0",
+    "check": "npm run audit && npm run deps",
```
koresar: Removing `npm outdated` since
* we have greenkeeper
* it exit code is 1 now on npm v4 :(

test/benchmark/.eslintrc
```diff
@@ -0,0 +1,8 @@
+// Add our overrides as necessary
+{
+  "rules": {
+    "global-require": 0,
```
koresar: Various relaxed rules to make sure it runs on bare bone node v4.

test/benchmark/index.js
```diff
@@ -0,0 +1,3 @@
+if (process.env.CI) {
```
koresar: Do not run tests if the `CI` env var is not set.

test/index.js
```diff
@@ -1,4 +1,4 @@
 require('require-all')({
   dirname: __dirname,
-  excludeDirs: process.env.CI ? /^\./ : /^\.|benchmark/
+  excludeDirs: /^\.|benchmark/
```
koresar: Do not run benchmark tests with the rest of the unit tests because it requires different runtime (node instead of babel-node).

koresar:
@FredyC ready to merge.

FredyC:
LGTM, thanks for handling it.

PR #228, opened 2016-08-25
allow easier usage by cjs modules

boneskull:
This modification alters how stampit is exported for Node.js.  Currently, if you are not using ES6 modules, you need to do this:

```js
const stampit = require('stampit').default;
const MyStamp = stampit();
```

This PR changes the appropriate dist file(s) so that the following is possible:

```js
const stampit = require('stampit');
const MyStamp = stampit();
```

There's `babel-plugin-add-module-exports`, but that assumes a `es2015-preset` is in use; it's not here.  I couldn't find a suitable Rollup plugin, so I just wrote my own.

Here's some REPL output of how it looks different:

Before:

```js
> require('stampit')
{ compose: [Function: compose],
  isComposable: [Function: isObject],
  isStamp: [Function: isStamp],
  methods: [Function: methods],
  properties: [Function: properties],
  refs: [Function: properties],
  props: [Function: properties],
  initializers: [Function: initializers],
  init: [Function: initializers],
  deepProperties: [Function: deepProperties],
  deepProps: [Function: deepProperties],
  staticProperties: [Function: staticProperties],
  statics: [Function: staticProperties],
  staticDeepProperties: [Function: staticDeepProperties],
  deepStatics: [Function: staticDeepProperties],
  configuration: [Function: configuration],
  conf: [Function: configuration],
  deepConfiguration: [Function: deepConfiguration],
  deepConf: [Function: deepConfiguration],
  propertyDescriptors: [Function: propertyDescriptors],
  staticPropertyDescriptors: [Function: staticPropertyDescriptors],
  default:
   { [Function: stampit]
     isStamp: [Function: isStamp],
     isComposable: [Function: isObject],
     compose: { [Function: _compose] staticProperties: [Object] },
     methods: [Function: methods],
     properties: [Function: properties],
     refs: [Function: properties],
     props: [Function: properties],
     initializers: [Function: initializers],
     init: [Function: initializers],
     deepProperties: [Function: deepProperties],
     deepProps: [Function: deepProperties],
     staticProperties: [Function: staticProperties],
     statics: [Function: staticProperties],
     staticDeepProperties: [Function: staticDeepProperties],
     deepStatics: [Function: staticDeepProperties],
     configuration: [Function: configuration],
     conf: [Function: configuration],
     deepConfiguration: [Function: deepConfiguration],
     deepConf: [Function: deepConfiguration],
     propertyDescriptors: [Function: propertyDescriptors],
[...]

PR #306, opened 2017-04-29
use Buble in tests; closes #305

### LINKED ISSUE #305: tests don't use same transpiler as dist bundles
The tests are bundled with Babel, and the dist files are bundled with Rollup and Buble.

This is a problem, because Babel does more stuff than Buble.  If, for example, Node v4 was missing `Array.prototype.includes`, and Babel transpiled it, but Buble ignored it, then we'd have false positives in CI, and Stampit would fail to run in Node v4.

Which is about where we're at (see `stampit.js:121`):

```
      if (isFunction(composer) && !uniqueComposers.includes(composer)) {
                                                   ^
TypeError: uniqueComposers.includes is not a function
```

Given nobody's actually complaining about this, I don't think `includes` needs to urgently be replaced with `indexOf`, so if there's a better way to fix the transpilation problem, maybe we should find it.

I have a PR forthcoming to run the tests via Buble + Rollup .  This was considerably more difficult than I had anticipated...

boneskull:
- add several rollup plugins to enable multiple entries (e.g. all test files)
- bundling happens in `build/index.js`
- add `reify` to dance around ES6 module loading
- output `dist/test.js` bundle of all tests
- remove unneeded `test/index.js`

also:

- add `build/index.js` to files covered by ESLint, and lint that file.
- remove "import is the same as require" test, because that's a test of the module loader, not stampit itself.  plus, it now fails due to reify.  ¯\_(ツ)_/¯


boneskull:
If I did this correctly, Travis CI should break on Node.js v4.

FredyC:
Well, it did break on v5 as well, is that intentional? :)

Anyway, this repo is kinda obsolete since there is a the monorepo now: https://github.com/stampit-org/stamp

koresar:
I'd need to test all that tomorrow. It's almost midnight here. 

Otherwise, @fredyC you can take the responsibility, test and merge it. I'd publish in 8 hours.

FredyC:
I am not sure it's a good idea to actually add a test bundle to `dist` folder which should really contain only files for distribution. 

Tests are actually failing on v5 as well because `Array.prototype.includes` is available only with `--harmony` flag there... check http://node.green/

As I've said in #305 I would replace this with `indexOf`, it's not worth it to use this feature at all cost.

Otherwise it looks good to me, so thanks for the effort @boneskull

package.json
```diff
@@ -38,25 +38,29 @@
...
   },
   "scripts": {
     "cov": "npm run cov:clean && npm run cov:generate",
     "cov:clean": "rimraf ./coverage/",
-    "cov:generate": "babel-node --presets=es2015 ./node_modules/.bin/istanbul cover test",
+    "cov:generate": "babel-node --presets=es2015 ./node_modules/.bin/istanbul cover ./dist/test.js",
     "prepublish": "npm run check",
     "pretest": "npm run build",
-    "test": "babel-node --presets=es2015 test && npm run lint",
+    "test": "node --require reify ./dist/test && npm run lint",
```
FredyC: Is it actually necessary to use reify? You are making a test bundle with CommonJS so it wont do anything here.

koresar:
Also, please check the size of the final bundle people. The min gzip better
be less than 2kb. It's 1.8kb atm.

FredyC:
@koresar These changes are not really touching dist bundles, it's about changes for tests only, so don't worry.

package.json
boneskull: Yes.

FredyC: Care to elaborate more ? :)

boneskull: The bundle is CJS but it doesn't matter since we don't export anything anyhow. But this would be the export of the bundle; it doesn't go thru and replace all the `import` statements in the tests.

boneskull:
I notice there are spread operators everywhere.  This also breaks Node.js v4, etc.

I am not really familiar with Buble/Rollup; the few times I've tried to use it, I couldn't make it do what I wanted.  I'm also unsure what prompted the switch from Babel...

boneskull:
We should publish no more than:

1.  A module consumable by Node.js v4 or greater.
2.  *Maybe* a browser-ready, CDN-able bundle, if that's something people actually need.
3.  The source files.

I'm going to assume others disagree.  I saw some rumblings about bad tooling...

FredyC:
Well, spread operator is a language feature. That should certainly transpiled by a Babel, however, es2015 preset does not include required plugins. Can you try switching to https://babeljs.io/docs/plugins/preset-env/ instead ?

I wouldn't really spend much more time in rewriting building & bundling, it's not worth it since we are going to throw this away eventually.

boneskull:
...but we're not using Babel here.  It appears Buble should take care of it, _but it doesn't_, and I don't know why.

boneskull:
(This isn't really about fussing with tooling; it's that any version of `stampit` published since the switch to Buble is likely broken in 4.x).

FredyC:
Ah I see, Buble does not support rest spread. I guess the easiest path is not to use this feature too. Can you handle it @boneskull please?

koresar:
Thanks for the ideas people.

There are few major issues this PR introduces.

The spread operator is already correctly transpiled. Whereas the `.includes()` is not. :(
See the distributable files: https://unpkg.com/stampit@3.1.2/dist/stampit.umd.js (UMD)
https://unpkg.com/stampit@3.1.2/dist/stampit.full.js (CJS)
The `.includes()` should be fixed asap.

@boneskull you removed a unit test, the src code for it you have implemented yourself about a year ago: https://github.com/stampit-org/stampit/pull/228/files#diff-dfbff627e6c248bcb3b61d7d06da9ca9R53
Removal of the unit test means that stampit v3 will no longer test support of the feature which some might rely on. This means, if we merge this PR we better release stampit v4.

This PR increases the complexity quite high. I take a chance to remind: if stampit was written in ES5 this PR would have never existed.

The unit tests still fail: https://travis-ci.org/stampit-org/stampit/jobs/227288366
What's the point to run tests in node.js v4 if test still fail in node.j v4? :)

I'm very reluctant taking these changes to the main codebase.

Sorry guys.

FredyC:
@koresar I don't see how this PR increases complexity really. If anything it is lowering it because using same transpiler solution for tests & source is something that should be taken seriously. The results of tests are somewhat useless if there is another layer. Can you elaborate more where is the complexity you talk about?

Unit tests fail simply because the use of rest spread which is not supported by Bublé. It's either about removing that feature from source or go other way around and bundle source with Babel again which I don't exactly remember why we have replaced. Leaving things as they are now is dangerous for sure.

koresar:
@FredyC true.

@boneskull could you please:
1) Return the test back.
2) Fix the unit tests to be actually runnable in node v4.
Thank you.

koresar:
The complexity I was referring to is the number of the `devDependencies`. It would be increased [by 4](https://github.com/stampit-org/stampit/pull/306/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R41). 24 total.

Again, if the stampit was written in ES5 the number of `devDependencies` would be 10.

boneskull:
My preferred solution would be to revert to using Babel. This will increase the number of dev deps. I'm not sure why that's a problem...

FredyC:
@boneskull Do you think it's a big problem to do those two points @koresar asked for?

I will be repeating myself, but since future of this repo is rather unsure, I don't think it's worth it to spend more time on it than it's necessary. Even if means ditching "rest&spread" from source.

boneskull:
@FredyC Yes. It's non-trivial using the existing tooling. I would have to use Babel because I don't know how to make Buble work.

boneskull:
There are also many places using spread parameters and it seems rather silly to me to replace them with whatever the ES5 alternative looks like.

FredyC:
Ok, let's close this then and worry about it when the real issue coming from different bundling solution arises.

PR #321, opened 2017-08-29
IE11 and node.js 0.10 support by adding 100 more bytes to the bundle

koresar:
This removes the burden many people have. They don't need to polyfil anything now. With this PR Stampit core feature - **dependencelessness** (What a word!) is preserved. The resulting minified bundle increases by 100 bytes, the .min.gz bundle increases by 40 bytes. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,13 +55,15 @@ $ bower install stampit
 or
 $ bower install stampit=https://npmcdn.com/stampit@3/dist/stampit.umd.min.js
 or
-$ bower install stampit=https://unpkg.com/stampit@3.1.2/dist/stampit.umd.js
+$ bower install stampit=https://unpkg.com/stampit@3.1.3/dist/stampit.umd.js
 ```
 
 Browsers: [![CDNJS](https://img.shields.io/cdnjs/v/stampit.svg)](https://cdnjs.com/libraries/stampit)
 [![UNPKG](https://img.shields.io/badge/unpkg.com--green.svg)](https://unpkg.com/stampit@latest/dist/stampit.umd.min.js)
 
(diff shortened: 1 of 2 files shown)


ericelliott:
IMO, [this is a bad idea](https://github.com/stampit-org/stampit/pull/320#issuecomment-325532589).

FredyC:
I think we can merge this for now and later we can try the approach through compat if needed. Question is, should we do something similar in `@stamp/it` also?

ericelliott:
Reminder: [we're adding ~10 Megabytes of wasted code to the world](https://github.com/stampit-org/stampit/pull/320#issuecomment-325812185) with this PR, and the correct solution for Stampit users is to [use a polyfill CDN](https://polyfill.io/v2/docs/).

It is not a library's responsibility to supply polyfills for application authors.

**Objection filed**. My work here is done. I wash my hands of this.

koresar:
@FredyC good question... Let me think.

README.md
```diff
@@ -55,13 +55,15 @@ $ bower install stampit
 or
 $ bower install stampit=https://npmcdn.com/stampit@3/dist/stampit.umd.min.js
 or
-$ bower install stampit=https://unpkg.com/stampit@3.1.2/dist/stampit.umd.js
+$ bower install stampit=https://unpkg.com/stampit@3.1.3/dist/stampit.umd.js
```
jyboudreau: Noticed this was published as 3.2.0 on NPM but docs here refering to 3.1.3. Probably not a big deal however.

ericelliott: Are we still maintaining the Bower version? Bower itself is no longer maintained.

koresar: Thanks. Will fix.

Issue #317, opened 2017-07-07
stampit___default.a(...).props is not a function in Webpack 3


jktravis:
Not sure if this is a problem with webpack or stampit, but I created an issue for it on the webpack repo. 

https://github.com/webpack/webpack/issues/5241

koresar:
Great bug reporting! Thank you 

I'm out of computer for the next few weeks. Can't run the example. 

Can you show full stack trace? Preferably with minification.

jktravis:
Sure. 

    TypeError: __WEBPACK_IMPORTED_MODULE_0_stampit___default.a(...).props is not a function
    at Object.defineProperty.value (http://localhost:63342/Webpack3StampIt/build/main.js:24:4)
    at __webpack_require__ (http://localhost:63342/Webpack3StampIt/build/commons.js:55:30)
    at Object. (http://localhost:63342/Webpack3StampIt/build/main.js:7:66)
    at __webpack_require__ (http://localhost:63342/Webpack3StampIt/build/commons.js:55:30)
    at webpackJsonpCallback (http://localhost:63342/Webpack3StampIt/build/commons.js:26:23)
    at http://localhost:63342/Webpack3StampIt/build/main.js:1:1

koresar:
Could you please check if it's reproducible with stampit@3.0.x ? 
There was a change in it which might have caused the issue.

If it still throws - then likely it's the webpack bug.
If it doesn't - the bug is likely ours.

jktravis:
It appears to still throw. I tried it with stampit@3.0.6

jktravis:
I received a reply on the webpack issue. Am I using it wrong now? 

https://github.com/webpack/webpack/issues/5241#issuecomment-313846314

koresar:
The provided workaround is correct. This appears to be stampit bug. I'll have some time to fix it in two weeks. 
Does the workaround work for you?

jktravis:
Thanks, @koresar. Yes. The workaround will do.

koresar:
Early to tell, but this might be fixed here #323

koresar:
Fixed with stampit v4.

Issue #304, opened 2017-04-28
Examples from readme


Mika83AC:
Hello everyone,

sorry, but am I just blind/an idiot/dumb or aren't the examples found in the readme working?

I installed stampit via bower via "bower install stampit" an have the following code:

```
<script src="bower_components\stampit\stampit.js"></script>
<script>
// Some privileged methods with some private data.
const Availability = stampit().init(function() {
  var isOpen = false; // private
  ....
```

When executing this in Chrome, it says:
Uncaught TypeError: stampit is not a function at stampit.html:5

But when typing "stampit" into Chrome Dev Tools console, it shows me that stampit is available with all it's methods (compose, composers, conf ...).

So what the hell is wrong?

Thx in advance!
Regards,
Michael

koresar:
Thank you for reporting.

It requires some investigation. Are you in the position to help?

FredyC:
@Mika83AC  Where do you have these script tags? Both in head? In case you include both at the same spot, chances are that your "app code" is executed before stampit script is loaded and parsed. You can investigate that in Network tab of developer tools.

The safest path is load external scripts in head tag and your logic at the end of body.

Mika83AC:
@koresar  Hm... I think I'm not in a position to help as I'm a JS and StampIt newbie ... :( Really have no idea what might be wrong.

@FredyC Even this is not working for me. Here is my example done as I understood your advice:

```
<!doctype html>
	<head>
		<meta charset="utf-8">
		
		<script src="bower_components\stampit\stampit.js"></script>
	</head>
	<body>
		Test
	</body>
	<script>
	// Some privileged methods with some private data.
	const Availability = stampit().init(function() {
	  var isOpen = false; // private

	  this.open = function open() {
		isOpen = true;
		return this;
	  };
	  this.close = function close() {
		isOpen = false;
		return this;
	  };
	  this.isOpen = function isOpenMethod() {
		return isOpen;
	  }
	});

	// Here's a stamp with public methods, and some state:
	const Membership = stampit({
	  methods: {
		add(member) {
		  this.members[member.name] = member;
		  return this;
		},
		getMember(name) {
		  return this.members[name];
		}
	  },
	  properties: {
		members: {}
	  }
	});

	// Let's set some defaults:
	const Defaults = stampit({
	  init({name, specials}) {
		this.name = name || this.name;
		this.specials = specials || this.specials;
	  },
	  properties: {
		name: 'The Saloon',
[...]

PR #326, opened 2017-10-21
Stampit v4

### LINKED ISSUE #323: Reducing size even more
Recently @ericelliott posted a truly great concern that we are bloating stampit: https://github.com/stampit-org/stampit/pull/320#issuecomment-325812185

I agree, we should save those bytes (lines of code). I would even add that we should simplify stampit infrastructure because it was overly complexified with ES6->ES5 compilation.

Current stampit UMD build size is `5.26 KB`. By removing fancy ES6 stuff (plus some optimization techniques) I managed to reduce the size down to `4.0 KB`.

To be precise: `5393` -> `4128` bytes. It is `1,265` bytes or 23.5% less.

Simultaneously, this will fix #317, which was reported three times IIRC. (TL;DR: our own examples do not work in browsers.)

Also, this will reduce `devDepencencies` a bit.

Pinging all people from the #320 discussion: @sethlivingston @FredyC @Mustack @jyboudreau @ericelliott 

Should I proceed and submit the PR?

koresar:
Closes #323 #317 #304 #68 

Changes:
* Removed the previously deprecated `refs`.
* Removed `stampit/*` utility functions. Advice to use `@stamp/*` modules instead.
* `composers` are not marked as experimental any more.
* Stampit becomes compatible with all `@stamp/*` stamps.
* The `.min.js` bundle size reduced by ~50%, the gzipped size reduced by ~40%.
* Tests can now be manually run in a browser (by opening `./test/index.html`)
* Stampit source code is ES5 now instead of ES6.
* Fixed perf bottleneck in node.js 8 (new V8 engine).

In case of silence I'll merge this PR in a week.


FredyC:
Phew, sorry @koresar, but I don't have a space to review this bulk of changes.

Honestly, I am not even sure why to bother with another stampit release when there is stamp monorepo. Why not to just have a `stampit@v4` to reference `@stamp/it` instead? Why two different code bases?

koresar:
1. Stampit will become compatible with monorepo. Some were complaining
about it.
2. Size reduce. Good for bragging and websites perf.

The other changes have just happened along the way.

FredyC:
Sorry, but that still does not explain why to keep two almost identical implementations. When I started the idea about monorepo I thought that eventually, stampit module will be just a redirect to this monorepo and after a while it might be even deprecated and removed.

Right now it's kinda unclear where is the difference. Is it possible to use extra modules from the monorepo with stampit? That's an example of a user question, I know they are both up to spec including composers, but in my opinion, it just makes things worse.

I can imagine it wasn't easy to accomplish size reduction you wanted, but it feels kinda OCD :) Do you know someone who actually complained that this lib is couple bytes larger than it could be? I am probably living in a different world where a size of the libs is so much higher and it does not really matter in overall.

Personally, I have kinda lost interest in stamps in overall. Perhaps I've never understood it that well to use it efficiently and without losing grasp over my codebase. In complex scenarios, it was just too messy for my taste. Not even mentioning adoption when working on the team and explaining everyone how it works.

If you feel like you want to go this direction then go ahead and merge and release it. I am not best to judge this anymore. Sorry friend.

koresar:
Stampit have to jump to next major version only because it is not following the latest spec v1.5.
If we would vote to make "composers" part of the spec year ago stampit would have stayed v3.
Unfortunately, community took a careful decision to do baby steps - implement "composers" as an experimental feature first. 
So, stampit v4 if a necessary last baby step towards compatibility with the latest specification v1.5.

AlexxNica:
@koresar Why did you remove the license?

license.txt
```diff
@@ -1,21 +0,0 @@
-The MIT License (MIT)
```
AlexxNica: Here

AlexxNica:
@koresar Little tip: You could have rebased your commits before pushing and merging to avoid polluting the history.

Take a look at:
https://git-scm.com/docs/git-rebase
https://git-scm.com/book/en/v2/Git-Branching-Rebasing

koresar:
Totally forgot to squash before merging. My bad.

Licence was removed by accident. Will readd, or feel free to pr. Sorry
about that.

PR #327, opened 2017-11-12
Remove references to out-of-date book and video.

ericelliott:


koresar:
LGTM.
What about this file?
https://github.com/stampit-org/stampit/blob/master/docs/pjabook-updated-examples.md

Should we leave it or remove too?

PR #338, opened 2018-09-15
Add another non-spec metadata - "name" to change factory name

koresar:
Proposal details: https://github.com/stampit-org/stamp-specification/issues/121

TL;DR:
```js
const HaveName = stampit({
  name: "Superman"
});

HaveName.name === "Superman"  // instead of the default "Stamp"
```

### DIFF
--- a/test/name.js
+++ b/test/name.js
@@ -47,6 +47,6 @@ try {
       t.end();
     });
   }
-} catch(e) {
+} catch (e) {
   // skip these tests in this environment
 }


koresar:
ping @FredyC

FredyC:
I don't know man, I never had this need for naming stamps really. It kinda feels that resulting name depends on the last named stamp composed in there. How can that be a good thing?

koresar:
Well, properties, initializer, methods, etc behave the same way.

> resulting [SOMETHING] depends on the last [HAS SOMETHING] stamp composed in there

Totally within the philosophy of stamps IMO.

koresar:
Here is a blog post draft: https://medium.com/@koresar/fun-with-stamps-episode-23-new-name-feature-4be91c27edcb

FredyC:
Ok sure, since this far from a breaking change, then go ahead :)

Btw, I am still unsure why you are actually maintaining this package when there is the `@stamp/it` one which you even introduced in your blog as _new way to do things_. You should probably do the same change there if you want them to be the same although I don't see the point really. I thought that eventually, the `stampit` will just a be a redirect to `@stamp/it` to have a sane codebase. Your call.

